PR #25, opened 2019-09-10
Move username, password, url and namespace to auth argument

aljazkosir:
Moved username, password, url and namespace under auth argument  as we talked about. Base doc fragment was changed so all other modules and tests needed some refactoring.

### DIFF
--- a/plugins/doc_fragments/base.py
+++ b/plugins/doc_fragments/base.py
@@ -10,27 +10,32 @@ __metaclass__ = type
 class ModuleDocFragment(object):
     DOCUMENTATION = """
 options:
-  user:
-    description:
-      - The username to use for connecting to the Sensu API.
-        If this is not set the value of the SENSU_USER environment variable will be checked.
-    type: str
-    default: admin
-  password:
-    description:
-      - The Sensu user's password.
-        If this is not set the value of the SENSU_PASSWORD environment variable will be checked.
-    type: str
-    default: P@ssw0rd!
-  url:
-    description:
-      - Location of the Sensu backend API.
-        If this is not set the value of the SENSU_BACKEND_URL environment variable will be checked.
-    type: str
-    default: http://localhost:8080
-  namespace:
-    description:
-      - RBAC namespace to operate in.
-    type: str
-    default: default
+  auth:
+    description: Authentication parameters. Can define each of them with ENV as well
+    type: dict
+    suboptions:
+      user:
+        description:
+          - The username to use for connecting to the Sensu API.
+            If this is not set the value of the SENSU_USER environment variable will be checked.
+        type: str
+        default: admin
+      password:
+        description:
+          - The Sensu user's password.
+            If this is not set the value of the SENSU_PASSWORD environment variable will be checked.
+        type: str
+        default: P@ssw0rd!
+      url:
+        description:
+          - Location of the Sensu backend API.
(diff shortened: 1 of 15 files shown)


plugins/doc_fragments/base.py
```diff
@@ -10,27 +10,31 @@
...
-        If this is not set the value of the SENSU_BACKEND_URL environment variable will be checked.
-    type: str
-    default: http://localhost:8080
-  namespace:
-    description:
-      - RBAC namespace to operate in.
-    type: str
-    default: default
+  auth:
+    description: Set of auth options
```
miha-plesko: "Authentication parameters. Can define each of them with ENV as well"

plugins/doc_fragments/base.py
```diff
@@ -10,27 +10,31 @@
...
+        default: P@ssw0rd!
+      url:
+        description:
+          - Location of the Sensu backend API.
+            If this is not set the value of the SENSU_BACKEND_URL environment variable will be checked.
+        type: str
+        default: http://localhost:8080
+      namespace:
+        description:
+          - RBAC namespace to operate in.
```
miha-plesko: Add
"If this is not set the value of the SENSU_NAMESPACE environment variable will be checked."

plugins/doc_fragments/base.py
```diff
@@ -10,27 +10,31 @@
...
+      password:
+        description:
+          - The Sensu user's password.
+            If this is not set the value of the SENSU_PASSWORD environment variable will be checked.
+        type: str
+        default: P@ssw0rd!
+      url:
+        description:
+          - Location of the Sensu backend API.
+            If this is not set the value of the SENSU_BACKEND_URL environment variable will be checked.
```
miha-plesko: Lets rename this to `SENSU_URL`

plugins/module_utils/arguments.py
```diff
@@ -12,22 +12,29 @@
...
+                fallback=(env_fallback, ["SENSU_USER"]),
+            ),
+            password=dict(
+                default="P@ssw0rd!",
+                no_log=True,
+                fallback=(env_fallback, ["SENSU_PASSWORD"]),
+            ),
+            url=dict(
+                default="http://localhost:8080",
+                fallback=(env_fallback, ["SENSU_BACKEND_URL"]),
```
miha-plesko: SENSU_URL

tests/integration/modules/molecule/sensu_go_check/playbook.yml
```diff
@@ -67,19 +70,19 @@
         env_vars:
           foo: bar
         runtime_assets: awesomeness
```
miha-plesko: Wait what, does this `awesomeness` asset really exist? If not, this then means we just add dummy values to our integration tests without actually creating the asset :) Besides, I thought it has to be array

aljazkosir: I don't think sensu API is checking asset existence when creating check. I just tried it locally with handler, it accepts anything you put there, but it will fail when the check is executed.

ansible list parameter also allows a single value (I also didn't know that), this is how they handle it :D

```
    if isinstance(value, list):
        return value

    if isinstance(value, string_types):
        return value.split(",")
    elif isinstance(value, int) or isinstance(value, float):
        return [str(value)]

    raise TypeError('%s cannot be converted to a list' % type(value))
```

PR #26, opened 2019-09-10
The sensu.sensu_go.install role

miha-plesko:
With this commit we introduce a role that configures target's package manager and then installs desired Sensu Go binary with it. Following binaries are currently present:

- sensu-go-backend
- sensu-go-agent
- sensu-go-cli

Currently the role supports following target platforms:

- Centos 6 & 7
- Ubuntu 14.04 & 16.04 & 18.04 & 18.10 & 19.04

Role allows you to either install latest version of the component or specify specific version, but that version must exist in the stable package repository.

### DIFF
--- a/roles/install/tasks/apt/install_one.yml
+++ b/roles/install/tasks/apt/install_one.yml
@@ -1,0 +1,17 @@
+---
+- name: Latest component version
+  set_fact:
+    pkg_name: '{{ item.name }}'
+    pkg_state: latest
+  when: item.version == 'latest'
+
+- name: Specific component version ({{ item.version }})
+  set_fact:
+    pkg_name: '{{ item.name }}={{ item.version }}-*'
+    pkg_state: present
+  when: item.version != 'latest'
+
+- name: Install component ({{ pkg_name }})
+  apt:
+    name: '{{ pkg_name }}'
+    state: '{{ pkg_state }}'
--- a/roles/install/tasks/yum/prepare.yml
+++ b/roles/install/tasks/yum/prepare.yml
@@ -1,0 +1,31 @@
+---
+- name: Include distro-specific vars ({{ ansible_distribution }})
+  include_vars: file='{{ ansible_distribution }}.yml'
+
+- name: Add yum repository
+  yum_repository:
+    name: sensu_stable
+    description: sensu_stable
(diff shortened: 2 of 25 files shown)


miha-plesko:
Hi @tadeboro a couple of side comments to give you some more context:

- role intentionally supports minimal number of parameters (just one) because we don't want user to get frustrated with everything configurable (like repo urls etc)
- role can be invoked manually, although usually it will be invoked as part of the higher-level roles (coming soon) that will also configure and start appropriate Sensu services
- molecule integration test consists of two parts (reflected in two Ansible groups - `latest_components` and `older_components`). On the former we install latest versions of all the packages. I think this test is the most important therefore we run it against all supported platforms. The other test is to see if role is able to install older version of the components and is not _that_ important - hence we only run it against single version of each OS type (centos, ubuntu)

roles/install/tasks/yum/prepare.yml
```diff
@@ -0,0 +1,31 @@
...
+    name: sensu_stable
+    description: sensu_stable
+    file: sensu
+    baseurl: https://packagecloud.io/sensu/stable/{{ os }}/{{ dist }}/$basearch
+    gpgkey: https://packagecloud.io/sensu/stable/gpgkey
+    gpgcheck: no
+    repo_gpgcheck: yes
+    enabled: yes
+    sslverify: yes
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
```
tadeboro: Is this line needed? I would assume that root certs are taken from here by default. But then again, I might be completely wrong.

roles/install/tasks/apt/install_one.yml
```diff
@@ -0,0 +1,17 @@
...
+  when: item.version == 'latest'
+
+- name: Specific component version ({{ item.version }})
+  set_fact:
+    pkg_name: '{{ item.name }}={{ item.version }}-*'
+    pkg_state: present
+  when: item.version != 'latest'
+
+- name: Install component ({{ pkg_name }})
+  apt:
```
tadeboro: I am pretty sure this command will also start the service on Ubuntu 14.04. Maybe we should set the `RUNLEVEL` env variable to `1` in order to keep the services down?

PR #27, opened 2019-09-11
Update asset module

aljazkosir:
With this commit we renamed and modify the return value of both `asset` and `asset_info` module.
I also changed the parameter from `download_url` to `url`, since `auth url` is now part of `auth` parameter.

\cc @miha-plesko


plugins/modules/asset.py
```diff
@@ -15,8 +15,11 @@
 }
 
 DOCUMENTATION = """
-module: sensu_go_asset
-author: "Cameron Hurst (@wakemaster39)"
+module: asset
+author:
+  - Aljaz Kosir (@aljazkosir)
```
miha-plesko: Don't drop Cameron, please keep him as first author in the list.

plugins/modules/asset.py
```diff
@@ -108,15 +125,15 @@ def main():
     client = arguments.get_sensu_client(module.params["auth"])
     path = "/assets/{0}".format(module.params["name"])
     payload = arguments.get_mutation_payload(
-        module.params, "download_url", "sha512", "filters", "headers",
+        module.params, "url", "sha512", "filters", "headers",
     )
-    payload["url"] = payload.pop("download_url")  # Remap download_url -> url
+    payload = get_api_payload(payload)
```
miha-plesko: It feels weird to have "get_mutation_payload" and "get_api_payload" separated. Couldn't we somehow unify this into a single function so that we get rid of "create payload then pop from it to build actual payload" because when you read the code it's like "okay so what was the initial 'payload' then, semi-payload" 😄 ?

Something like:

```python
def build_api_payload(params):
  payload = arguments.get_mutation_payload(params)  # NOTICE: no flat parameters here, so we won't need to pop them later
  payload["builds"] = [dict(
    url=params.get("url"),
    sha512=params.get("sha512"),
    ...
  )]
  return payload
```

plugins/modules/asset_info.py
```diff
@@ -15,8 +15,11 @@
 }
 
 DOCUMENTATION = """
-module: sensu_go_asset_info
-author: "Paul Arthur (@flowerysong)"
+module: asset_info
+author:
+  - Aljaz Kosir (@aljazkosir)
```
miha-plesko: Same here, please never drop the original author from the list. List him first

aljazkosir:
@miha-plesko reverted, as we talked about.

plugins/modules/asset_info.py
miha-plesko: @aljazkosir it's Paul Arthur here

miha-plesko:
@tadeboro would you like to have some saying here? I'm about to merge

tadeboro:
I have nothing to add, @aljazkosir did a great job. So go smash that merge button.

PR #28, opened 2019-09-12
Update check module

aljazkosir:
Updated check and check info.


tests/integration/modules/molecule/check/playbook.yml
```diff
@@ -5,8 +5,54 @@
   hosts: all
   gather_facts: true
   tasks:
+    - name: Create a check with missing required parameters parameters
```
miha-plesko:
```suggestion
    - name: Create a check with missing required parameters
```

plugins/modules/check.py
```diff
@@ -126,58 +130,71 @@
...
+    payload = arguments.get_mutation_payload(
+        params, 'command', 'subscriptions', 'handlers', 'interval', 'cron', 'publish', 'timeout',
+        'ttl', 'stdin', 'low_flap_threshold', 'high_flap_threshold', 'proxy_entity_name',
+        'output_metric_format', 'round_robin'
+    )
+    payload['check_hooks'] = utils.dict_to_single_item_dicts(params['check_hooks'])
+    payload['env_vars'] = utils.dict_to_key_value_strings(params['env_vars'])
 
-        self.path = '/checks/{0}'.format(self.params['name'])
+    # Special cases when even if it's empty, None is still returned from server
```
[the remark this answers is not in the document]
miha-plesko: UPDATE: We should ensure similar behavior for all other non-required parameters as well. If parameter is **not required** and user did **not** explicitly set value for it, then we must not include it in payload at all.

tests/unit/modules/test_check.py
```diff
@@ -0,0 +1,148 @@
...
+        sync_mock = mocker.patch.object(utils, "sync")
+        sync_mock.return_value = True, {}
+        set_module_args(
+            name="test_check",
+            command='echo "test"',
+            subscriptions=['switches'],
+            interval=60
+        )
+
+        with pytest.raises(AnsibleExitJson) as context:
```
miha-plesko:
```suggestion
        with pytest.raises(AnsibleExitJson):
```

plugins/modules/check.py
```diff
@@ -126,58 +130,71 @@
...
-                self.payload[key] = self.params[key]
+    # Apparently empty string is also a valid choice and it's the default
+    payload['output_metric_format'] = payload.pop('output_metric_format', '')
 
-        self.param_dict_to_payload_list('check_hooks')
-        self.param_dict_to_payload_kv_list('env_vars')
+    if params['proxy_entity_attributes']:
+        payload['proxy_requests'] = {
+            'entity_attributes': params['proxy_entity_attributes'],
+            'splay': params['proxy_splay'],
```
miha-plesko: "splay" value is not a required parameter. In case user didn't set "proxy_splay" explicitly we shouldn't be sending that key at all. Sorry, I know it complicates code a little :)

tests/unit/modules/test_check.py
```diff
@@ -0,0 +1,148 @@
...
+            high_flap_threshold=0,
+            low_flap_threshold=0,
+            round_robin=False,
+            stdin=False,
+            runtime_assets=None,
+            metadata=dict(
+                name="test_check",
+                namespace="default",
+            ),
+        )
```
miha-plesko: The payload above should really be:

```python
dict(
  command='echo "test"',
  subscriptions=['switches'],
  interval=60,
  metadata=dict(
    name="test_check",
    namespace="default",
  )
)
```

All other default values are wrong - they should not be part of the payload. Please do following:

- update module args specification so that it doesn't set default values for any of optional parameters
- update the code so that it doesn't include the parameter in payload when the value is None

tests/unit/modules/test_check.py
```diff
@@ -0,0 +1,135 @@
...
+            proxy_requests=dict(
+                entity_attributes=['entity.entity_class == "proxy"'],
+                splay=True,
+                splay_coverage=90
+            ),
+            output_metric_format='nagios_perfdata',
+            output_metric_handlers=['influx-db'],
+            round_robin=True,
+            env_vars=dict(foo='bar'),
+            runtime_assets='awesomeness'
```
miha-plesko: It's a bit funny that we specify so much arguments here, because we raise anyway. I'd go with minimal set of params :)

PR #33, opened 2019-09-15
Introduce daily test

tadeboro:
Changes in this PR add daily integration tests for master and stable branches. It contains quite a few changes because we had to refactor the CircleCI configuration file to allow for future extensions with different python and Ansible versions.

For detailed description, see individual commit messages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,33 @@
 version: "2.1"
 
+workflows:
+  version: 2
+  main_workflow:
+    jobs:
+      - sanity
+
+      - unit_python27_ansible29
+      - unit_python35_ansible29
+      - unit_python36_ansible29
+      - unit_python37_ansible29
+
+      - integration_ansible29_git
+
+# A list of all supported python versions.
 executors:
   python27:
     docker: [ { image: circleci/python:2.7 } ]
     environment: { PY: "2.7" }
+    working_directory: ~/ansible_collections/sensu/sensu_go
+
+  python35:
+    docker: [ { image: circleci/python:3.5 } ]
+    environment: { PY: "3.5" }
+    working_directory: ~/ansible_collections/sensu/sensu_go
+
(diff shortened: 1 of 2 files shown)


tests/integration/modules/molecule/asset/molecule.yml
```diff
@@ -2,6 +2,15 @@
 scenario:
   name: asset
 platforms:
-  - name: instance
-    image: sensu/sensu:latest
-    command: sensu-backend start --state-dir /var/lib/sensu/sensu-backend/etcd1 --state-dir /var/lib/sensu/sensu-backend/etcd1 --log-level debug
+  - name: v5.13.1
```
miha-plesko: Wouldn't we want to test against `latest`, too? Doing so our daily tests would "detect" incompatible version the same day it's released by Sensu. Currently we would need to manually update the version here to see if the new version works as expected.

.circleci/config.yml
```diff
@@ -1,17 +1,151 @@
...
+      - schedule:
+          cron: "12 5 * * *"
+          filters: { branches: { only: [ master ] } }
+    jobs:
+      - integration_ansible29_git
+
+  daily_released:
+    triggers:
+      - schedule:
+          cron: "21 5 * * *"
```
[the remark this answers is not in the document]
tadeboro: I scheduled them at different times just to be nicer to CircleCI backend. The hardest part of the test is setting up a dedicated worker and this is usually done in minutes. All the things we do later on are relatively light on CPU.

But sure, I will schedule them one hour apart as suggested.

tests/integration/modules/molecule/asset/molecule.yml
tadeboro: I was thinking a bit about this, and testing against nightly would make sense since we would be able to catch regressions before release. But as you said, testing against the latest release should be better than nothing.

PR #34, opened 2019-09-15
Bump version to 0.7.0

tadeboro:
The version bump is a bit larger than one would expect because we broke the end-user API.

### DIFF
--- a/galaxy.yml
+++ b/galaxy.yml
@@ -1,6 +1,6 @@
 namespace: sensu
 name: sensu_go
-version: 0.3.0
+version: 0.7.0
 
 authors:
   - Paul Arthur <paul.arthur@flowerysong.com> (@flowerysong)


galaxy.yml
```diff
@@ -1,6 +1,6 @@
 namespace: sensu
 name: sensu_go
-version: 0.3.0
+version: 0.7.0
```
miha-plesko: I'm fine with `0.7.0`, but out of curiosity, why by 4 version points? :) Wouldn't it be more obvious to bump to `0.10.0`? Perhaps even `1.0.0` (although I know we're not stable yet)

tadeboro: To be completely honest, we have put version 0.7.2 on AnsibleFest presentation slides quite some time ago. And this is the first step towards getting repository and slides in sync ;)

As for the 1.0.0, I would wait a bit with it. Since we have quite some modules to check, we might need to adjust public API a bit. And I would rather change the API in the 0.x versions that by semver do not guarantee anything about API stability.

miha-plesko: Okay, merging

Issue #20, opened 2019-08-21
Consider splitting some modules


miha-plesko:
There are some modules that should probably get split into more modules in order to keep them simple and with manageable set of input parameters. For example the

```
sensu_go_handler
```
([link](https://github.com/sensu/sensu-go-ansible/blob/master/plugins/modules/sensu_go_handler.py))

module. There are 3 major handler types available as per https://docs.sensu.io/sensu-go/5.11/reference/handlers/#how-do-sensu-handlers-work:

- pipe handlers
- TCP/UDP handlers
- handler set

Each of those major types supports slightly different set of input parameters. So it seems more user-friendly to provide three modules to do the job:

```
sensu_go_pipe_handler
sensu_go_socker_handler
sensu_go_handler_set
```

Handler module is just one example of a currently too complex module. There is also `sensu_go_check` module with two or three dozens of input parameters that should probably be somehow split as well, if possible...

PR #35, opened 2019-09-16
Add pipe handler module

aljazkosir:
We are splitting handler module into 3 separate modules, `pipe_handler` is one of them.

Addressing issue: https://github.com/sensu/sensu-go-ansible/issues/20


plugins/modules/handler_info.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2019, Paul Arthur <paul.arthur@flowerysong.com>
```
miha-plesko: Every file you make is our own, so please remove Paul from here (and from the documentation below)

tests/integration/modules/molecule/pipe_handler/Dockerfile.j2
```diff
@@ -0,0 +1,14 @@
+# Molecule managed
```
miha-plesko: This should be a symlink pointing to `../shared/Dockerfile.j2`

tests/integration/modules/molecule/pipe_handler/molecule.yml
```diff
@@ -0,0 +1,7 @@
+---
+scenario:
+  name: pipe_handler
+platforms:
+  - name: instance
+    image: sensu/sensu:latest
+    command: sensu-backend start --state-dir /var/lib/sensu/sensu-backend/etcd1 --state-dir /var/lib/sensu/sensu-backend/etcd1 --log-level debug
```
miha-plesko: Two things: 
- `--state-dir` is used twice
- we could use multi-line format

```
command: > 
  sensu-backend start
    --state-dir /var/lib/sensu/sensu-backend/etcd1
    --log-level debug
```

tests/integration/modules/molecule/pipe_handler/playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+        auth:
+          url: http://localhost:8080
+        name: minimal_handler
+        command: sensu-influxdb-handler
+      register: result
+
+    - assert:
+        that:
+          - result is changed
+          - result.object.metadata.name == 'minimal_handler'
```
miha-plesko: Also assert the command

tests/integration/modules/molecule/pipe_handler/playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+        auth:
+          url: http://localhost:8080
+        name: minimal_handler
+        command: sensu-influxdb-handler
+      register: result
+
+    - assert:
+        that: result is not changed
+
+    - name: Create an pipe handler
```
miha-plesko: an -> a

tests/integration/modules/molecule/pipe_handler/playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+        env_vars:
+          INFLUXDB_ADDR: http://influxdb.default.svc.cluster.local:8086
+          INFLUXDB_USER: sensu
+          INFLUXDB_PASS: password
+        runtime_assets:
+          - sensu-influxdb-handler
+      register: result
+
+    - assert:
+        that: result is changed
```
miha-plesko: I think we should add more assert statements to check all the parameters; unless there is some problem doing it I don't see?

tests/integration/modules/molecule/pipe_handler/playbook.yml
```diff
@@ -0,0 +1,129 @@
...
+          - result.objects.0.metadata.name == 'handler'
+
+    - name: Delete a pipe handler
+      pipe_handler:
+        auth:
+          url: http://localhost:8080
+        name: handler
+        state: absent
+
+    - name: Get all pipe handler
```
miha-plesko: handlers

tests/integration/modules/molecule/pipe_handler/Dockerfile.j2
aljazkosir: Hmm every time I run molecule, new Dockerfile.j2 is generated and it overwrites the symlink

aljazkosir: Actually it doesn't overwrite the symlink... my bad. Molecule fails to bring up docker containers if the file is missing.

miha-plesko: Huh, you got me on this one @aljazkosir because the symlink never got overwritten in my PC and I was wondering if I was doing it wrong :) For now, please just convert it to symlink. In the future, molecule will probably become smart enough to use default for this file (ahem, @tadeboro, when "in the future" depends on you I'm afraid 😄 )

tests/integration/modules/molecule/pipe_handler/molecule.yml
```diff
@@ -0,0 +1,10 @@
+---
+scenario:
+  name: pipe_handler
+platforms:
```
miha-plesko: I think we ought to test against three platforms (5.12.0, 5.13.1 and latest), please see how @tadeboro modified it for asset module

tests/integration/modules/molecule/pipe_handler/playbook.yml
```diff
@@ -0,0 +1,140 @@
+---
+- name: Converge
+  collections:
+    - sensu.sensu_go
+  hosts: all
+  gather_facts: true
```
miha-plesko: You probably copy-pasted this, but we could actually set this to `no` because we're just accessing the REST API, and gathering facts is time consuming. Let me open issue that we must fix it for all modules; integration tests will run quite faster I think

miha-plesko: UPDATE: actually, leave it for now, we can do a followup, I opened an issue for it

tests/integration/modules/molecule/pipe_handler/molecule.yml
aljazkosir: oh yeah, totally forgot about that

PR #40, opened 2019-09-18
Add handler set module

aljazkosir:
Last handler in the line. With this commit we introduce `handler_set`.

Closes: #20


miha-plesko:
@aljazkosir please go with `handler_set` here. I know it's a bit different than `socket_handler` and `pipe_handler`, but "set_handler" sounds just ridiculous :)

PR #43, opened 2019-09-19
Update filter module

aljazkosir:
Update old `sensu_go_filter` to the new module structure.


tests/unit/modules/test_filter_info.py
```diff
@@ -0,0 +1,47 @@
...
+    errors, utils,
+)
+from ansible_collections.sensu.sensu_go.plugins.modules import filter_info
+
+from .common.utils import (
+    AnsibleExitJson, AnsibleFailJson, ModuleTestCase, set_module_args,
+)
+
+
+class TestilterInfo(ModuleTestCase):
```
miha-plesko:
```suggestion
class TestFilterInfo(ModuleTestCase):
```

Found a nitpick 👏 🎆 🍾

aljazkosir: oops :)

PR #50, opened 2019-09-23
Add filter example

aljazkosir:
I just saw that I forgot to add an example to filter module.

### DIFF
--- a/plugins/modules/filter.py
+++ b/plugins/modules/filter.py
@@ -44,6 +44,38 @@ options:
 '''
 
 EXAMPLES = '''
+- name: Create a filter
+  filter:
+    name: filter
+    action: deny
+    expressions:
+      - event.check.interval == 10
+      - event.check.occurrences == 1
+    runtime_assets: awesomeness
+
+- name: Create a production filter
+  filter:
+    name: filter
+    action: allow
+    expressions:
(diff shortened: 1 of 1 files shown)


plugins/modules/filter.py
```diff
@@ -44,6 +44,14 @@
...
 
 EXAMPLES = '''
+- name: Create a filter
+  filter:
+    name: filter
+    action: deny
+    expressions:
+      - event.check.interval == 10
+      - event.check.occurrences == 1
+    runtime_assets: awesomeness
```
miha-plesko: It's nice to have one simple example like the one above, but it would also be great to provide at least one "real world" example from https://docs.sensu.io/sensu-go/5.13/reference/filters/ , for example this one https://docs.sensu.io/sensu-go/5.13/reference/filters/#built-in-filter-only-incidents

aljazkosir: But this is actually a pipe handler that uses built-in filter `is_incident`. :sweat_smile:

miha-plesko: I see, the 2nd link is wrong. I actually meant this one https://docs.sensu.io/sensu-go/5.13/reference/filters/#handling-repeated-events

aljazkosir: I added this one and few other examples.

PR #53, opened 2019-09-25
Update hook module

aljazkosir:
Updated hook module to the new module structure. 

Sensu documentation says that timeout is not required, however when you make a PUT request on their API, you get a 400 error response if `command` and `timeout` are missing.


plugins/modules/hook.py
```diff
@@ -0,0 +1,120 @@
...
+  hook:
+    auth:
+      url: http://localhost:8080
+    name: restart_nginx
+    command: sudo systemctl start nginx
+    timeout: 60
+    stdin: false
+
+- name: Capture the process tree
+
```
miha-plesko: Strange why this time sanity didn't complain

PR #54, opened 2019-10-03
[WIP] Update event module

aljazkosir:
Pheew, we finally did it! :)
This module required some changes in our base functions. Because there are nested dicts in the payload we needed a custom comparator, however this change is also useful for previous modules so we could replace the old comparator with the one I defined now, in a follow-up PR of course, WDYT?

Event api had some strange responses and some of the parameters mentioned on their API are not even writable.

\cc @miha-plesko


plugins/module_utils/arguments.py
```diff
@@ -57,10 +57,14 @@
 )
 
 
-def get_mutation_payload(source, *wanted_params):
-    payload = {
+def get_spec_payload(source, *wanted_params):
```
tadeboro: Can you add a few simple tests for this function? I know we are testing it indirectly already, but I would like to see some tests just for the sake of explaining what we expect this function to do.

plugins/modules/event.py
```diff
@@ -0,0 +1,221 @@
...
+                    'status': ['preview'],
+                    'supported_by': 'XLAB Steampunk'}
+
+DOCUMENTATION = '''
+module: event
+author:
+  - Paul Arthur (@flowerysong)
+  - Aljaz Kosir (@aljazkosir)
+  - Miha Plesko (@miha-plesko)
+  - Tadej Borovsak (@tadeboro)
```
tadeboro: I wrote zero lines of code in this module and do not deserve to be on this list, so feel free to remove me from it.

plugins/modules/event.py
```diff
@@ -0,0 +1,221 @@
...
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    arguments, errors, utils, debug
+)
+
+
+def do_differ(current, desired):
```
tadeboro: Should we add a test or two to the module unit test?

Also, we probably need to decide how to handle structural mismatches. The original comparison function reports those as differences, while custom implementation for this module will raise an exception.

I guess the right way to go would be to detect those mismatches and report them to the user. Most of the time, those mismatches will indicate that Sensu Go made a backward-incompatible change to the API. And we would like to get notified about them with as much information as possible. Any suggestions on how to approach this?

tests/integration/modules/molecule/event/playbook.yml
```diff
@@ -0,0 +1,210 @@
...
+      event:
+        auth:
+          url: http://localhost:8080
+        timestamp: 134532453
+        entity: "{{ entity.object }}"
+        check:
+          metadata:
+            name: not_so_awesome_check
+            namespace: default
+          interval: 60
```
miha-plesko: @aljazkosir I was expecting following format:

```yaml
check: "{{ check.object }}"
check_merge:
  interval: 60
```

Wouldn't it be more user-friendly this way?

plugins/modules/event.py
miha-plesko: Originally we wanted to use
```
- Steampunk
```
in there, but I think sanity complains unless you specify a valid GH handle next to it. Then we decided to just list the three related engineers instead. Maybe we should create a GH user named "steampunk" to handle situations like this?

plugins/module_utils/utils.py
```diff
@@ -9,7 +9,18 @@
 from ansible_collections.sensu.sensu_go.plugins.module_utils import errors
 
 
-def sync(state, client, path, payload, check_mode):
+def do_differ(current, desired):
```
miha-plesko: So by default (here) you use do_differ function that only verifies the 1st level of keys. For some specific modules (event) you implement an enhanced function that goes into depth to check all nested fields as well.

=> I suggest you use the enhanced function by default, not just for the event module

plugins/modules/event.py
aljazkosir: Yes of course, I will add some test.

As for the structural changes, in the case of structural mismatch we could throw an error explaining what when wrong and abort the module. Continuing the module is probably not wise since the structure between local and remote is different, and with that kind of API we don't really know what would happen :).
Our integration tests should notify us about the change.

For example if the remote payload is:

```
{"a": "b", "c": "d"}
```
and local is:
```
{"a": "b", "c": {"d": "e"}}
```

In this case we would abort the module and report the error to user with something like this:
`"Remote and local value on key 'c' structurally mismatch"`.

plugins/module_utils/utils.py
aljazkosir: I tried this and tests were all passing, but I wanted a second opinion on this :smile: . Will apply that change.

plugins/modules/event.py
miha-plesko: Been thinking about this for 10min now, good question :)

The UX we currently offer with our modules does not bring the concept of "diff" to user yet - she doesn't really know that we're even performing GET request for her (prior PUTing). She just knows we're creating entity as she specified - and I think this should be maintained. Following that UX I'm not sure we would want to report structural mismatch back to the user.

How can we debug it then, in case of error?? Well, recently we introduced simple logging mechanism that logs all REST API requests and all responses. In case of backward-incompatible change we would turn this logging on and manually spot the difference. Or, if we don't have access to her environment, she would send the log file to us to examine.

Bottomline, my suggestion is to silently "fail" i.e. to mark the entity as "changed" and proceed to PUT request. In case of incompatible API change the PUT will fail (with 400) and at that point user will reach out to us. Then we bring out the debug-logger procedure. Looking forward to your opinion.

tadeboro: I tend to agree with @miha-plesko here: we should make sure that the comparison does not raise an exception on structural mismatches and let the thing error-out later down the line. But I would still add a debug message if we detect the issue when comparing the payloads. This should make it a bit easier to determine the root cause of the issue down the line.

plugins/modules/event.py
tadeboro: I would avoid creating a new GitHub user just for this since someone would need to monitor the messages on that artificial thing.

plugins/module_utils/utils.py
```diff
@@ -31,12 +31,24 @@ def sync(state, client, path, payload, check_mode):
     return False, remote_object
 
 
+def is_different_type(a, b):
```
miha-plesko: Could be private method

plugins/module_utils/utils.py
```diff
@@ -31,12 +31,24 @@ def sync(state, client, path, payload, check_mode):
...
+    if isinstance(a, string_types) and isinstance(b, string_types):
+        return False
+    return (a and b) and not isinstance(a, type(b))
```
miha-plesko: Why do we need that `(a and b)` part?

plugins/modules/event.py
```diff
@@ -0,0 +1,207 @@
...
+    auth:
+      url: http://localhost:8080
+    name: awesome_check
+    command: echo "Check command"
+    output_metric_handlers:
+      - influxdb
+    subscriptions:
+      - checks
+      - also_checks
+    interval: 30
```
miha-plesko: Are these parameters all required? We should go as simple as possible with other entities (and focus on event module)

plugins/modules/event.py
```diff
@@ -0,0 +1,207 @@
...
+      - Entity object returned from M(entity) or M(entity_info).
+    type: dict
+  check:
+    description:
+      - Check object returned from M(check) or M(check_info).
+    type: dict
+  check_merge:
+    description:
+      - Information about the status and history of the event.
+        U(https://docs.sensu.io/sensu-go/5.13/reference/events/#check-attributes)
```
miha-plesko:
```suggestion
        U(https://docs.sensu.io/sensu-go/latest/reference/events/#check-attributes)
```

plugins/modules/event.py
```diff
@@ -0,0 +1,207 @@
...
+  register: check
+
+- name: Create event with minimal parameters
+  event:
+    auth:
+      url: http://localhost:8080
+    entity: "{{ entity.object }}"
+    check: "{{ check.object }}"
+    check_merge:
+      output: Check command
```
miha-plesko:
```suggestion
      output: Hello World
```

or something similar. If we put "Check command" it feels like bash command is supposed to be passed here, not the output of it

plugins/modules/event.py
```diff
@@ -0,0 +1,207 @@
...
+'''
+
+RETURN = '''
+objects:
```
miha-plesko:
```suggestion
object:
```

tests/integration/modules/molecule/event/playbook.yml
```diff
@@ -0,0 +1,212 @@
...
+        auth:
+          url: http://localhost:8080
+        name: awesome_check
+        command: /bin/true
+        output_metric_handlers:
+          - influxdb
+        subscriptions:
+          - checks
+          - also_checks
+        interval: 30
```
miha-plesko: Same here, if those parameters are not required then we better omit them

tadeboro:
When I started the second round of review, something felt weird about the whole thing. So I went to check the Sensu Go's documentation on events again, searching for things that we might have missed when we designed the API for the first time. What follows is a summary of the upstream documentation and some proposed changes to the module API.
An essential part of the event description is an entity that is the originator of the event in question. All other fields are optional and carry additional pieces of information about the event's nature: associated check, check results, metrics, and timestamp.

With all this in mind, I would write the playbook task like this:

```
- event:
    entity: my_entity

    timestamp: 1234567
    check:
      name: my_check
      duration: 1.234
      executed: 12345678
      ...
    metrics:
      handlers: [ bla ]
      points: [ p1, p2 ]
```

I deliberately left out any `entity` and `check` data that is not strictly related to the event since those two should already exist before event insertion.

Since this API is purelly imperative, I would not bother with implementing the info module at all. If the Sensu guys can explain what `GET /events/entity/check` actually returns (latest procesed event, something else, ...), maybe we can add info module then, but for now, this makes little to no sense.

aljazkosir:
@tadeboro a lot of docs up there are misleading for this object. For example, metrics parameter is not even saved on the API, it just gets left out. Event and check don't need to exists beforehand and check's interval is actually required parameter. Same applies for some of the entity parameters.

The proposed playbook was actually the one we had in mind when writing the module, but after debugging numerous frustrating 400 code errors we found out that the API is very different from the docs.
That is why we have a `check_merge` which merges check definition and check response dictionaries. And metrics were left out since they are not really parsed. I should write this in the description, sorry about that.

tadeboro:
@aljazkosir Did we talk to the Sensu people about this? I would be very much interested in what the general idea behind the event API is. Even more, without that information, creating a sensible Ansible module will be next to impossible.

plugins/module_utils/utils.py
aljazkosir: I only check type for non None types, because while it is a different type, it's not really a structural mismatch, it just means that that value is empty, which happens quite often when reading data form remote.

miha-plesko:
We had a discussion with @aljazkosir about what the purpose of the API endpoint might be and in the end we think it's something like:

*For a given **entity** we report an "occurance" of a specific **check**. Occurance can have some extra data given, like exact command output, or not. Sensu then stores last 20 occurances*

I'm not sure if we want to play too smart here with the module - if there is the REST API for check (and there is) there is also the module.

plugins/module_utils/utils.py
miha-plesko: I see, can you do it explicitly then, please, so it'll read better:

```python
if a is None or b is None:
  return False
else:
  return not isinstance(a, type(b))
```

aljazkosir: I renamed the function to `is_same_type`. I think it's a bit easier to read and we lose all the negating in the function.

aljazkosir:
Closed in favor of https://github.com/sensu/sensu-go-ansible/pull/104

PR #57, opened 2019-10-09
Update namespace module

aljazkosir:
Updated namespace module.
There is also endpoint `/api/enterprise/user-namespaces`, whcih should return namesapces user has access to. But after investigating and assigning different permissions to the user, API always returned empty array.


tadeboro:
I would leave the `enterprise` API endpoint for now, since the name suggests that we will need an active subscription if we would like to test it.

I saw that we create completely new auth section documentation for namespace modules since the namespace parameter is not needed here. Maybe we can still reuse the old document fragments and argument spec and force that parameter to `None` in the module itself? Then we can simply add documentation note that `namespace` auth parameter is ignored and be done with it.

Among other things, this will make our playbooks more forgiving and malleable, since end-users can now reuse the same auth block in all tasks. If we tighten the spec, this is no longer true since Ansible will complain about unknown parameter.

What do you think @aljazkosir?

aljazkosir:
@tadeboro Makes sense. We can also use the extended argument spec that includes `name`, `state`, `labels` and `annotations`. In documentation we add note that `labels` and `annotations` are also ignored.

plugins/modules/namespace_info.py
```diff
@@ -0,0 +1,70 @@
...
+  - Miha Plesko (@miha-plesko)
+  - Tadej Borovsak (@tadeboro)
+short_description: Lists Sensu namespaces
+description:
+  - For more information, refer to the Sensu documentation at
+    U(https://docs.sensu.io/sensu-go/latest/reference/rbac/#namespaces)
+extends_documentation_fragment:
+  - sensu.sensu_go.base
+notes:
+  - Parameters C(namespace) is ignored in this module.
```
tadeboro: Can we fix this one too, please? I promise I will merge it then ;)

aljazkosir: Ops, I totally forgot about the _info module :sweat_smile:.

PR #58, opened 2019-10-10
Initial implementation of the sensu.sensu_go.backend role

miha-plesko:
The role is supposed to be run against nodes where we want to have the Sensu backend running. It installs the sensu-backend there (using the `install` role which is also part of this same collection) then it constructs the /etc/sensu/backend.yml configuration file based on the vars. Eventually it starts backend as a system service (systemctl or whatever the target OS supports).

Basic integration tests (with default values for parameters) are run against all supported target platforms. Later, when we enhance the role with more advanced configuration options, I would suggest, yes, that we add new scenarios, but it's sufficient that they are run against one platform only.


miha-plesko:
A comment on integration tests: we're using a proxy container because CircleCI's main CI/CD container (the executor) is not connected to the same network as the molecule-created containers. To workaround it, we let molecule provision the proxy container then we use

```
docker exec proxy "curl <BACKEND_CONTAINER>"
```

Only this way we're able to actually perform some HTTP requests to the resulting Sensu Backend containers to test the REST API.

PR #61, opened 2019-10-13
Update user module

aljazkosir:
With this PR we introduce user module. 

Module supports user creation and modification. 
Sensu API does not support removal of user through DELETE request, it only disables the user.
User password is required for every modification, the  only exception is DELETE request, where the only requirement is username.

EDIT: It just crossed my mind that maybe we could remove the support of DELETE request and only allow end-user to change `disable` status of user through PUT requests? WDYT @tadeboro


plugins/modules/user.py
```diff
@@ -0,0 +1,153 @@
...
+        if check_mode:
+            return True, payload
+        utils.put(client, path, payload)
+        return True, utils.get(client, path)
+
+    return False, remote_object
+
+
+def is_password_required(remote_object, state):
+    return not (state == 'disabled' and remote_object is not None)
```
tadeboro: If we apply the de Morgan's law here, we can convert the condition into

    state != "disabled" or remote_object is None

Both parts of this condition are a bit counter-intuitive. I guess we cannot do much about the first part since this condition seems to be directly linked to the API requirements. But maybe we can do something about the second part?

I am a bit worried that our end-users will find it a bit counter-intuitive if they need to specify a password when disabling a non-existing backend user. Maybe we should create a disabled user only if the `password` parameter has been set? If we decide to go this way, having a task

    sensu.sensu_go.user:
      name: Lizzy
      state: disabled

in the playbook would not create new user `Lizzy` if she does not exist yet. On the other hand, putting a task

    sensu.sensu_go.user:
      name: Lizzy
      state: disabled
      password: pass

would create a new (disabled) user Lizzy with the password set to `pass`. Any opinions about this approach?

plugins/modules/user.py
```diff
@@ -0,0 +1,153 @@
...
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    arguments, errors, utils,
+)
+
+
+def do_differ(current, desired):
```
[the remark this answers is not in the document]
aljazkosir: Now I'm really wondering why I wrote this comparator :sweat_smile: . And yes your assumption is correct, API omits the password field.

plugins/modules/user.py
aljazkosir: I like it, and we can remove the `is_password_required` entirely and only use ansible's `required_if`. If user tries to disable a non existent user we stop the module and tell him that he cannot do that.

tadeboro:
@aljazkosir I like what I see in the latest iteration of the code. I will do the full review first thing in the morning.

tests/integration/modules/molecule/user/playbook.yml
```diff
@@ -0,0 +1,118 @@
...
+      user_info:
+        auth:
+          url: http://localhost:8080
+        name: test_username
+      register: result
+
+    - assert:
+        that:
+          - result.objects.0.username == 'test_username'
+          - result.objects.0.disabled == True
```
tadeboro: Would it be possible to add a test that makes sure we fail if we try to disable the non-existing user? I was thinking about adding something like this:

    - name: Try to disable non-existent user
      user:
        auth:
          url: http://localhost:8080
        name: missing_user
        state: disabled
      ignore_errors: true
      register: result

    - assert:
        that:
          - result.failed

Note that I have not tested if this works at all, I am just being a smartass here ;)

tests/integration/modules/molecule/user/playbook.yml
```diff
@@ -0,0 +1,118 @@
...
+        name: test_username
+        password: hidden_password?
+        state: disabled
+        groups:
+          - dev
+          - prod
+      register: result
+
+    - debug:
+        msg: "{{result}}"
```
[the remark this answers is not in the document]
aljazkosir: Oopsie :sweat_smile:

tests/integration/modules/molecule/user/playbook.yml
aljazkosir: Yes, I'll add the test. 
BTW the test you provided worked without any modifications :smiley:

PR #62, opened 2019-10-15
Add initial support for backend config

matejart:
This expands the list of supported variables in the role, which translate into the Sensu backend's configuration file.

This commit addresses the configurations of Sensu backend and Sensu agent, the API and dashboard. I skipped the settings that I consider managed and the subject of future commits: caching, KPI-related settings, and etcd store settings.

References:
  * https://docs.sensu.io/sensu-go/5.14/reference/backend/#configuration
  * https://docs.sensu.io/sensu-go/5.14/files/backend.yml

### DIFF
--- a/roles/backend/templates/backend.yml.j2
+++ b/roles/backend/templates/backend.yml.j2
@@ -3,17 +3,8 @@
 # {{ managed }}
 #
 
-##
-# Sensu backend configuration
-##
-{# -------------------------------------- #}
-{% if backend_debug %}
-debug: {{ backend_debug }}
-{% endif %}
-{# -------------------------------------- #}
-{% if backend_log_level %}
-log-level: {{ backend_log_level }}
-{% endif %}
-{# -------------------------------------- #}
 state-dir: "/var/lib/sensu/sensu-backend"
-{# -------------------------------------- #}
+
+{% if backend_config %}
+{{ backend_config | to_nice_yaml }}
+{% endif %}
--- a/roles/backend/README.md
+++ b/roles/backend/README.md
@@ -11,16 +11,51 @@ This role has no requirements.
 Role Variables
 --------------
 
-Most variables are directly interpolated into the /etc/sensu/backend.yml configuration file
-which is documented in the [official Sensu documentation][backend-conf]. Please refer to it to
-learn what exactly each specific parameter does.
+The role uses the `backend_config` top-level variable to populate the Sensu
+backend's `/etc/sensu/backend.yml`configuration file as is documented in the
+[official Sensu documentation][backend-conf]. The following `backend_config.*`
+variables can be used in the playbooks:
 
-| Variable            | Default Value | Description |
-|---------------------|---------------|-------------|
-| backend_debug       | /             | see [Sensu docs][backend-conf]
-| backend_log_level   | /             | see [Sensu docs][backend-conf]
+| Variable    | Examples  | Description |
+|-------------|-----------|-------------|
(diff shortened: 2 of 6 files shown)


roles/backend/templates/backend.yml.j2
```diff
@@ -16,4 +16,48 @@ log-level: {{ backend_log_level }}
 {% endif %}
 {# -------------------------------------- #}
 state-dir: "/var/lib/sensu/sensu-backend"
+config-file: "/etc/sensu/backend.yml"
```
tadeboro: Maybe we should group all backend-related variables into a single variable `backend_config` and loop over the contents here? Alternatively, we can expose the `backend_config` to the user and document what options are available. (I am more partial to the second solution because it is more straightforward.)

If we group the variables, the whole template reduces to:

```
{% for name, value in backend_config.items() %}
{%   if value %}
{{ name }}: {{ value }}
{%   endif %}
{% endfor %}
```

matejart: Thank you for these suggestions. I decided to go for both, so we now mostly generate the contents of the `/etc/sensu/backend.yml` from the `backend_config` items.

roles/backend/templates/backend.yml.j2
```diff
@@ -3,17 +3,9 @@
...
-debug: {{ backend_debug }}
-{% endif %}
-{# -------------------------------------- #}
-{% if backend_log_level %}
-log-level: {{ backend_log_level }}
-{% endif %}
-{# -------------------------------------- #}
 state-dir: "/var/lib/sensu/sensu-backend"
-{# -------------------------------------- #}
+config-file: "/etc/sensu/backend.yml"
```
tadeboro: Does this line make any difference? Once it is interpreted, Sensu already finished reading the configuration and I would be surprised to see it restart the configuration process.

If we would like to support using a custom configuration file, we should probably pass it as a command-line parameter. But this means that the service that is responsible for controlling the Sensu backend must support passing custom flags to the executable.

I would say that we remove this line and be done with it.

roles/backend/templates/backend.yml.j2
```diff
@@ -3,17 +3,9 @@
...
-{# -------------------------------------- #}
-{% if backend_debug %}
-debug: {{ backend_debug }}
-{% endif %}
-{# -------------------------------------- #}
-{% if backend_log_level %}
-log-level: {{ backend_log_level }}
-{% endif %}
-{# -------------------------------------- #}
 state-dir: "/var/lib/sensu/sensu-backend"
```
tadeboro: I know this line is not part of the current changeset, but I would rather see `/var/lib/sensu/sensu-backend` be the _default_ state directory and allow users to override it if they need to. We can probably solve this with a simple conditional:

```
{% if "state-dir" not in backend_config %}
state-dir: "/var/lib/sensu/sensu-backend"
{% endif %}
```

But feel free to solve this issue in a separate follow-up PR.

roles/backend/templates/backend.yml.j2
matejart: 👍 
I've left it in because somehow I thought it was funny in how it self-references, even in their example config file. I have not thought of an adverse effect of the deamon restarting (indefinitely?).

roles/backend/templates/backend.yml.j2
matejart: 👍 
Even though this puts us back to the if-land, I suppose it's not that bad to have just one-off instead of coming up with an elaborate way of baking into the vars the defaults.

PR #63, opened 2019-10-17
Address special case backend configs

matejart:
We now allow for a user change of the configurations that were previously baked into the template: `state-dir`.

### DIFF
--- a/roles/backend/templates/backend.yml.j2
+++ b/roles/backend/templates/backend.yml.j2
@@ -3,7 +3,9 @@
 # {{ managed }}
 #
 
+{% if not backend_config or "state-dir" not in backend_config %}
 state-dir: "/var/lib/sensu/sensu-backend"
+{% endif %}
 
 {% if backend_config %}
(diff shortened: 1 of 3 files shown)


roles/backend/templates/backend.yml.j2
```diff
@@ -3,7 +3,9 @@
 # {{ managed }}
 #
 
+{% if not backend_config or not backend_config["state-dir"] %}
```
tadeboro: Does this conditional survives this:

    vars:
      backend_config:
        debug: yes
    roles:
      - sensu.sensu_go.backend

I would expect the second part of the conditional to explode. If this is the case, maybe we should use `"state-dir" not in backend_config`?

matejart: Thanks for spotting this issue. I have updated the code and also added another piece of tests.

matejart:
Once CircleCI green-lights the commits, they should be ready for final review.

PR #67, opened 2019-10-20
Introduce Sphinx documetation builder

tadeboro:
Having the best Ansible collection on Earth with no documentation is not much better than having no Ansible collection at all. Changes that we introduced in the current pull request bootstrap our documentation structure.

Although the changeset is relatively small, it can already do quite a lot. And the best way to test it is to run this sequence of commands:

    $ pip install -r docs.requirements
    $ make docs

The first command will install the required Python packages, and the second one will render the documentation. Assuming everything went smoothly, we can find our freshly generated documentation in the `docs/build/html` folder. To view the documentation, open the `docs/build/html/index.html` file with a web browser.


Makefile
```diff
@@ -0,0 +1,32 @@
+# Make sure we have ansible_collections/sensu/sensu_go as a prefix. This is
+# ugly as heck, but it works. I suggest to all future developer to treat next
+# few lines as an opportunity to learn a thing or two about GNU make ;)
+collection := $(notdir $(realpath $(CURDIR)      ))
+namespace  := $(notdir $(realpath $(CURDIR)/..   ))
+toplevel   := $(notdir $(realpath $(CURDIR)/../..))
+
+err_msg := Place collection at <WHATEVER>/ansible_collections/sensu/sengo_go
+ifneq (sensu_go,$(collection))
+  $(error $(err_msg)
```
mancabizjak: We're missing the closing `)` here, resulting in  `Makefile:12: *** unterminated call to function 'error': missing ')'.` in case the collection resides at an invalid path.

Same thing on L12 and L14 as well.

docs/Makefile.custom
```diff
@@ -0,0 +1,18 @@
...
+modules := $(shell find ../plugins/modules/ -name '*.py' | grep -v sensu_go_)
+module_docs := $(modules:../plugins/modules/%.py=source/modules/%.rst)
+
+.PHONY: all
+docs: $(module_docs)
+	$(MAKE) html
+
+.PHONY: clean
+clean:
+	rm -rf build source/modules/*
```
mancabizjak: This leaves us with an empty `source/modules` directory. Which leaves me wondering, why aren't we deleting `source/modules` altogether?

```suggestion
	rm -rf build source/modules
```

tadeboro: When I initially wrote this, I had no `mkdir -p source/modules` command in the documentation extraction rule. Because `ansible-doc-extractor` expects target directory to exist, this was an easy way of making sure clean does not break the build.

Of course, I changed a lot of things since then and forgot about this rule. So thanks for noticing my inconsistency.

Makefile
```diff
@@ -0,0 +1,32 @@
+# Make sure we have ansible_collections/sensu/sensu_go as a prefix. This is
+# ugly as heck, but it works. I suggest all future developer to treat next few
+# lines as an opportunity to learn a thing or two about GNU make ;)
+collection := $(notdir $(realpath $(CURDIR)      ))
+namespace  := $(notdir $(realpath $(CURDIR)/..   ))
+toplevel   := $(notdir $(realpath $(CURDIR)/../..))
+
+err_msg := Place collection at <WHATEVER>/ansible_collections/sensu/sengo_go
```
aljazkosir: `sengo_go` -> `sensu_go`

PR #69, opened 2019-10-21
Split CI pipeline into stages

tadeboro:
Up until now we ran all of our jobs in parallel. And while this sounds great in theory, it wastes quite a few resources when one of the short-running tests fail.

This is why we split the CI jobs into two categories: fast and slow ones. Now we first run the fast ones and then proceed to run slow ones. Note that the slow jobs are only run if all of the fast ones finished successfully.

This saves a ton if resources if one of the fast tests fails.

/cc @mancabizjak

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,14 +11,26 @@ workflows:
       - unit_python36_ansible29
       - unit_python37_ansible29
 
-      - modules_integration_ansible29_git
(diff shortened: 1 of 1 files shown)


tadeboro:
This is how our workflow looks after this change:

Screenshot_2019-10-21 CircleCI - CircleCI

aljazkosir:
Thanks, @tadeboro

PR #70, opened 2019-10-22
Update role and role info modules

mancabizjak:
With this PR we port `role` and `role_info` modules to the new module structure. 

Both modules are now scoped to namespaced roles only. Support for clustered roles (operating across namespaces) will be added separately, in dedicated modules, and is thus omitted here.


plugins/modules/role_info.py
```diff
@@ -0,0 +1,78 @@
...
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
```
aljazkosir: XLAB Steampunk

tests/integration/modules/molecule/role/Dockerfile.j2
```diff
@@ -0,0 +1,14 @@
+# Molecule managed
```
aljazkosir: This should be a symlink pointing to `../shared/Dockerfile.j2`

plugins/modules/role_info.py
```diff
@@ -0,0 +1,78 @@
...
+module: role_info
+author:
+  - Paul Arthur (@flowerysong)
+  - Manca Bizjak (@mancabizjak)
+  - Aljaz Kosir (@aljazkosir)
+  - Tadej Borovsak (@tadeboro)
+short_description: Lists Sensu roles
+description:
+  - For more information, refer to the Sensu documentation at
+    U(https://docs.sensu.io/sensu-go/latest/reference/roles/)
```
aljazkosir: 404 page not found :smiley:, I think you meant `https://docs.sensu.io/sensu-go/latest/reference/rbac/`

plugins/modules/role.py
```diff
@@ -0,0 +1,135 @@
...
+      resources:
+        description:
+          - Types of resources the rule has permission to access.
+        type: list
+        required: yes
+      resource_names:
+        description:
+          - Names of specific resources the rule has permission to access.
+        type: list
+        default: []
```
tadeboro: Is there a reason why `resouece_names` has a default value set? There is nothing wrong with it. I am just curious where `[]` comes from. But more importantly, the argument spec below does not have this default value set.

plugins/modules/role.py
```diff
@@ -0,0 +1,135 @@
...
+  name:
+    description:
+      - The Sensu object's name.
+    type: str
+    required: yes
+  state:
+    description:
+      - Target state of the Sensu object.
+    type: str
+    choices: [ "present", "absent" ]
```
tadeboro: Nitpick: YAML does not need quotes here. Feel free to ignore it.

tests/integration/modules/molecule/role/Dockerfile.j2
mancabizjak: Oops, I totally missed this. Done, thanks :+1:

plugins/modules/role_info.py
mancabizjak: :scream: Fixed

tests/unit/modules/test_role.py
```diff
@@ -0,0 +1,141 @@
...
+                    resources=[],
+                )
+            ],
+        )
+        with pytest.raises(AnsibleFailJson):
+            role.main()
+
+    def test_failure_invalid_verb(self, mocker):
+        sync_mock = mocker.patch.object(utils, 'sync')
+        sync_mock.return_value = True, {}
```
tadeboro: Do we need those two lines? I would expect that parameter validation kicks in before we get busy interacting with the API.

plugins/modules/role.py
```diff
@@ -0,0 +1,141 @@
...
+    type: dict
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    arguments, errors, utils
+)
+
+
+def validate_module_params(module):
```
tadeboro: I turned out that we do not need this check. To quote Sensu people:

> Resource names are not considered when enforcing RBAC for create verb. I'm checking with our engineering team for their take.

> So you can create a rule that contains a resource_names and the create verb, but resource_names will not be taken into account on create (POST) requests. You could however use the update verb (which translates to PUT requests) with resource_names, which effectively grants permissions to create and update a resource via PUT requests.

So all we need to do document this and let the users do whatever they like.

tests/unit/modules/test_role.py
tadeboro: I was so happy that I found copy-paste remnant that I completely forgot to think about anything else ;) And now that I know why those two lines are here, I would like to suggest a minor adaptation. Instead of setting a return value in `sync` mock object, can we instead specify a side effect and raise an exception? For example

    sync_mock.side_effect = Exception("Validation should fail before we get to sync")

The main reason for my plead is: I am getting old and I will forget this discussion in abour an hour or so. And when the test fails in the future, I am sure the error message in that exception will be the most valuable thing ever ;)

PR #71, opened 2019-10-22
Update silence and silence_info modules

aljazkosir:
With this PR we update sensu go `silence` and `silence_info` modules. API of this module ignores the `metadata.name` parameter because it is autogenerated from `subscription` and `check` in the form of `$SUBSCRIPTION:$CHECK` we can use this form as an input to the module's `name`. 

So for example if end-user wants to silence all the checks on a specific subscription, he can call the module like so:
```
- name: Silence all checks on a specific subscription
  silence:
    name: proxy:*
```

He can also target specific entity by taking advantage of per-entity subscriptions:
```
- name: Silence all checks on a specific entity
  silence:
    name: entity:important-entity:*
```

I couldn't decide between a module having two parameters (`subscription` and `check`) which are then merged to generate `name` used for url, or module with one parameter (`name`) which is split into two parts, `subscription` and `check`. In the end I decided to use only one parameter (`name`). I will be happy to hear your opinion on this @tadeboro and @mancabizjak.

Closes: https://github.com/sensu/sensu-go-ansible/issues/78


tadeboro:
A quick glimpse at the [official docs](https://docs.sensu.io/sensu-go/5.10/reference/silencing/#spec-attributes) where `subscription` and `check` are separate parameters suggests that we should probably do the same. Another thing that bothers me with the composite name is how to properly split the name into two parts if there is more than one `:` in it. I imagine the answer is *the rightmost colon wins*, but I would rather not guess if I do not have to.

aljazkosir:
@tadeboro I updated to module so we have two parameters, which are used to generate `name`. I also split our arguments into more variables, maybe it's a bit of a overkil, but it allows us to reuse those variables in special cases lke this.

There is also an issue with `name` parameter. API says it's not required but in realitiy it is, it must match the name that is autogenerated from the subscription and check

plugins/modules/silence.py
```diff
@@ -0,0 +1,162 @@
...
+
+    client = arguments.get_sensu_client(module.params['auth'])
+    path = '/silenced/{0}'.format(name)
+    module.params['name'] = name
+    payload = arguments.get_mutation_payload(
+        module.params, 'subscription', 'check', 'begin', 'expire', 'expire_on_resolve', 'reason'
+    )
+
+    try:
+        changed, asset = utils.sync(
```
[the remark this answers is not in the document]
aljazkosir: Ooops, you got me :sweat_smile:

plugins/modules/silence.py
```diff
@@ -0,0 +1,162 @@
...
+                type='int',
+            ),
+            expire_on_resolve=dict(
+                type='bool'
+            ),
+            reason=dict()
+        ),
+    )
+
+    name = '{0}:{1}'.format(module.params['subscription'], module.params['check'])
```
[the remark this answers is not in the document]
aljazkosir: Yep, not valid. We should not allow it

plugins/modules/silence.py
```diff
@@ -0,0 +1,161 @@
...
+# Copyright: (c) 2019, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'XLAB community'}
```
mancabizjak: `XLAB community` -> `XLAB Steampunk`. It's funny how we keep messing this up :grinning:

plugins/modules/silence_info.py
```diff
@@ -0,0 +1,98 @@
...
+# Copyright: (c) 2019, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'XLAB community'}
```
mancabizjak: Same here, `XLAB Steampunk`

tadeboro:
@aljazkosir Can you please rebase your work on the `master` branch? Doing this will allow you to drop most of the argument-related mess from the changeset.

plugins/modules/silence.py
aljazkosir: Ooops :D

plugins/modules/silence_info.py
```diff
@@ -0,0 +1,95 @@
...
+  description: list of Sensu silence entries
+  returned: always
+  type: list
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    arguments, errors, utils
+)
+from ansible_collections.sensu.sensu_go.plugins.modules.silence import name_is_valid
```
tadeboro: This will not work. Ansiballz framework only bundles things from the `module_utils` package.

aljazkosir: oh... of course :sweat_smile:

aljazkosir:
@tadeboro hmm, yes I think we should treat them as equal. Or maybe even better we should remove the `defaullt: *`. this way comparison between such payloads would not be needed. If `subscription` or `check` is None, asterisk (*) only appears in the name, not in the actuall fields. WDYT?

tadeboro:
Removing default values would be my preferred way of handling this, but I did not want to force you doing this in this PR. But if you feel like fixing it hare and now, I am all for it.

plugins/modules/silence.py
```diff
@@ -0,0 +1,143 @@
...
+    U(https://docs.sensu.io/sensu-go/latest/reference/silencing/)
+extends_documentation_fragment:
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.state
+  - sensu.sensu_go.labels
+  - sensu.sensu_go.annotations
+options:
+  subscription:
+    description:
+      - "The name of the subscription the entry should match. If left empty a silencing entry will
```
mancabizjak: You can remove the quotes here, just to be consistent across the doc

plugins/modules/silence.py
```diff
@@ -0,0 +1,143 @@
...
+            expire=dict(
+                type='int',
+            ),
+            expire_on_resolve=dict(
+                type='bool'
+            ),
+            reason=dict()
+        ),
+    )
+    if module.params['subscription'] == '*' and module.params['check'] == '*':
```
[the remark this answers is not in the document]
mancabizjak: Maybe you could assign `module.params['subscription']` and  `module.params['check']` to `subscription` and `check` variables, to make it a bit easier to read later on:
```suggestion
    subscription = module.params['subscription']
    check = module.params['check']
    if subscription == '*' and check == '*':
        module.fail_json(msg="One of 'subscription' or 'check' must contain a name")

    name = '{0}:{1}'.format(subscription, check)
```
But I think I like @tadeboro's suggestion better, as it is much more concise and removes the need for these additional assignments anyways.

plugins/modules/silence.py
aljazkosir: `ansible-test` sanity checker complains if there is a colon followed by a space. I reorganized the text so `:` is not needed anymore.

aljazkosir:
I removed the default values as discussed and added few more tests.

plugins/modules/silence.py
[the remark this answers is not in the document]
aljazkosir: This section of the code was removed, because now that default values are gone, ansible's `required_one_of` is doing this job for us :)

PR #72, opened 2019-10-22
Add support for agent config

matejart:
Developing support for creating `/etc/sensu/agent.yml`.


matejart:
@tadeboro I just need to add 9 more variable descriptions, but I'll go ahead and mark this PR as reviewable. The CircleCI should be green soon.

roles/agent/README.md
```diff
@@ -11,16 +11,52 @@ This role has no requirements.
...
-
-| Variable            | Default Value | Description |
-|---------------------|---------------|-------------|
-| agent_backends      | /             | see [Sensu docs][agent-conf]
-| agent_log_level     | /             | see [Sensu docs][agent-conf]
-
-[backend-conf]: https://docs.sensu.io/sensu-go/latest/reference/agent/#configuration-summary
+| Variable            | Examples  | Description |
+|---------------------|-----------|-------------|
+| agent_backend_urls  | `["ws://0.0.0.0:8081", "ws://0.0.0.0:8082"]` | ws or wss URL of the Sensu backend server |
```
tadeboro: I think we should treat this variable as an implementation detail since its only purpose is to allow the automatic generation of `backend-url` from the inventory file. If the user wants to set the backend urls manually, she can do so by adding `backend-url` config into `agent_config` variable.

Issue #59, opened 2019-10-10
Get documentation fragments and common argument specifications back in sync


tadeboro:
At the moment, common argument specifications in `plugins/module_utils/arguments.py` and documentation in `plugins/doc_fragments/*.py` are out of sync a bit. While not being much of a problem right now, getting them back in sync would allow us to remove some of the duplicated documentation from our modules.

Note that this should probably be fixed after we convert all of the remaining `sensu_go_` prefixed modules in order to avoid sanity check having a mental breakdown.

tadeboro:
This has been fixed by #73

PR #73, opened 2019-10-23
Make argument spec composable

tadeboro:
The primary purpose of this PR is twofold. First, we tried to make modules self-documenting. And secondly, we tried to increase the malleability of the shared parameter specification and documentation.

We tried to achieve this by removing the opaque groups of shared parameters and replace them with a function call that lists what argument specifications we need. We did roughly the same thing with documentation and split it into fragments that document each share parameter separately.


aljazkosir:
Nice, a better version of what I tried to do :), will change https://github.com/sensu/sensu-go-ansible/pull/71 when we merge this.

This should also resolve our https://github.com/sensu/sensu-go-ansible/issues/59 issue, I think.

PR #83, opened 2019-10-25
Update role_binding and role_binding_info modules

mancabizjak:
Addresses #76.

In this PR we update two modules: `role_binding`, and `role_binding_info`. 

Following the scoping of recently updated `role` and `role_info` modules (#70), role binding modules also deal with namespaced roles only, leaving cluster role bindings for future extraction into a separate module.

To allow for intuitive use of the `role_binding` module, we split `subjects` parameter (containing a mixture of users and groups to bind to a role) expected by the Sensu backend, to `users` and `groups` parameters. Custom validation logic was added to ensure that at least one of these is provided when creating a role binding.

Following the discussion in #88, I added more thorough idempotence checks when creating role bindings. To ensure that `role_binding` doesn't change the state for an existing role binding with subjects provided in a different order, I added another check in `module_utils.do_differ`. I'm not particularly fond of the implementation as I wasn't sure what was the best way to approach this, considering that we expect changes in `do_differ` for `role` and `cluster_role` in any case. Looking forward to your opinions @tadeboro @aljazkosir.


plugins/module_utils/utils.py
```diff
@@ -4,6 +4,7 @@
 # GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
 
 from __future__ import absolute_import, division, print_function
+
```
tadeboro: Nitpick: We do not need this ;)

plugins/module_utils/utils.py
```diff
@@ -36,12 +37,26 @@ def do_differ(current, desired):
...
 
     for key, value in desired.items():
-        if value != current.get(key):
+        current_value = current.get(key)
+        # list elements are of the form {"type": "a", "name": "b"}
+        if (isinstance(value, list) and
+                all([isinstance(i, dict) and "name" in i and "type" in i for i in value])):
+            if type_name_dicts_differ(value, current_value):
+                return True
+        elif value != current_value:
```
tadeboro: I think that we should leave keep the default comparator as simple as possible. Instead of extending this comparator to handle all use cases, we can create specialized ones and then pass them to the `sync` function as an additional argument (that we need to add, of course ;).

Sync method declaration would look something like this:

    def sync(state, client, path, payload, check_mode, comparator=do_differ):

We can then hide all role-binding-specific stuff next to the main function or in a separate role utils module. This last one might be a better approach since cluster role bindings might need similar functionality.

Does this sound sane?

plugins/module_utils/utils.py
```diff
@@ -36,12 +37,26 @@ def do_differ(current, desired):
...
 
 
+# sorts a list of dicts by "type" and "name" keys
+# and returns the result of comparison.
+def type_name_dicts_differ(a, b):
+    sorted_a = sorted(a, key=lambda x: (x["type"], x["name"]))
+    sorted_b = sorted(b, key=lambda x: (x["type"], x["name"]))
+    return sorted_a != sorted_b
+
+
```
tadeboro: Should we move this to a separate role utils module (alongside with the custom comparator that I was fantasizing about)?

plugins/module_utils/utils.py
```diff
@@ -83,3 +98,10 @@ def dict_to_single_item_dicts(data):
 
 def dict_to_key_value_strings(data):
     return ["{0}={1}".format(k, v) for k, v in data.items()]
+
+
+def to_type_name_dict(type, name):
+    return {
+        "type": type,
+        "name": name,
+    }
```
tadeboro: Again, should we move this is a separate role utils module?

plugins/modules/role_binding.py
```diff
@@ -0,0 +1,133 @@
...
+)
+
+
+def validate_module_params(module):
+    params = module.params
+    if params["state"] == "present":
+        if not (params["users"] or params["groups"]):
+            module.fail_json(
+                msg="missing required parameters: users or groups"
+            )
```
tadeboro: I like this a lot.

aljazkosir:
Maybe we could create a custom `do_differ` function and pass it to the `utils.sync` function. `utils.sync`'s default comparator would still remain the same. This way we can avoid having one giant comparator.

plugins/module_utils/utils.py
aljazkosir: Separate role utils module might come in handy for `role` and `cluster_role` special comparators too.

mancabizjak: I agree, thanks!

mancabizjak:
I applied some changes, most importantly I moved role-binding-specific utility functions to `role_utils` module and made some simplifications to utility functions themselves, compared to the previous version. 

Following @aljazkosir's good example, I added a test suite for the newly introduced utility module. Our `role_binding` module now provides a custom `role_utils.do_differ` comparator to `sync`.

mancabizjak:
Yep, it looks exactly the same indeed :slightly_smiling_face: I imagined we would merge the conditionals in a shared `do_differ`. Considering that we will have (cluster)role and (cluster)role-binding utilities all packed in `role_utils`, it didn't seem so bad to have a common `do_differ` as well.

On second thought, having `role_bindings_do_differ` and `roles_do_differ` seems much cleaner as it wouldn't mix the specifics of two different modules in a single comparator function. That feels kind of wrong and defeats the purpose of truly custom comparators. So I think renaming is a better option :+1:

tests/unit/modules/test_role_binding.py
```diff
@@ -0,0 +1,149 @@
...
+                    type='Group',
+                ),
+                dict(
+                    name='user_1',
+                    type='User',
+                ),
+                dict(
+                    name='user_2',
+                    type='User',
+                ),
```
tadeboro: Maybe we should compare this the same way we compare subjects in code? Because currently, if the sensu API decides to spice things up and start returning subjects in arbitrary order, our test will fall flat on its face.

tadeboro:
> should I split into 2 commits to have the change of definition of `utils.sync` recorded in commit history (saw the comment in #88)?

Am I having a bit of trouble keeping on top of those two PRs. They are just too similar and I keep losing track of what should be done. So please, do split it, and thank you for reminding me about this.

mancabizjak:
Done :+1:

PR #88, opened 2019-10-30
Update cluster_role and cluster_role_info modules

aljazkosir:
This PR is a follow-up of https://github.com/sensu/sensu-go-ansible/pull/70. We add cluster role and cluster role info support as a new modules. Cluster roles adds support for `Cluster-wide resource types` in addition to namespaced resource types.

### DIFF
--- a/plugins/module_utils/role_utils.py
+++ b/plugins/module_utils/role_utils.py
@@ -42,3 +42,36 @@ def _do_subjects_differ(a, b):
     sorted_a = sorted(a, key=lambda x: (x['type'], x['name']))
     sorted_b = sorted(b, key=lambda x: (x['type'], x['name']))
     return sorted_a != sorted_b
+
+
+def _rule_set(rules):
+    return {
+        (
+            frozenset(r.get('verbs', []) or []),
+            frozenset(r.get('resources', []) or []),
+            frozenset(r.get('resource_names', []) or [])
+        ) for r in rules
+    }
+
+
+def _do_rules_differ(current_rules, desired_rules):
+    if len(current_rules) != len(desired_rules):
+        return True
+    if _rule_set(current_rules) != _rule_set(desired_rules):
+        return True
+    return False
+
+
+def do_roles_differ(current, desired):
+    if current is None:
+        return True
+
+    for key, value in desired.items():
+        current_value = current.get(key)
+        if key == 'rules':
+            if _do_rules_differ(current_value, value):
+                return True
+        elif value != current.get(key):
+            return True
+
+    return False
--- a/plugins/modules/cluster_role.py
+++ b/plugins/modules/cluster_role.py
@@ -1,0 +1,137 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2019, Paul Arthur <paul.arthur@flowerysong.com>
+# Copyright: (c) 2019, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'XLAB Steampunk'}
+
+DOCUMENTATION = '''
+module: cluster_role
+author:
+ - Paul Arthur (@flowerysong)
+ - Manca Bizjak (@mancabizjak)
+ - Aljaz Kosir (@aljazkosir)
+ - Tadej Borovsak (@tadeboro)
+short_description: Manages Sensu cluster roles
+description:
+  - For more information, refer to the Sensu documentation at
+    U(https://docs.sensu.io/sensu-go/latest/reference/rbac/)
+extends_documentation_fragment:
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.name
+  - sensu.sensu_go.state
+notes:
+  - Parameter C(auth.namespace) is ignored in this module.
+options:
+  rules:
+    description:
+      - Rules that the cluster role applies.
+    type: list
+    suboptions:
+      verbs:
+        description:
+          - Permissions to be applied by the rule.
+        type: list
+        required: yes
+        choices: [get, list, create, update, delete]
+      resources:
+        description:
+          - Types of resources the rule has permission to access.
+        type: list
+        required: yes
+      resource_names:
+        description:
+          - Names of specific resources the rule has permission to access.
+          - Note that for the 'create' verb, this argument will not be
+            taken into account when enforcing RBAC, even if it is provided.
+        type: list
+'''
+
+EXAMPLES = '''
+- name: Create a cluster role
+  cluster_role:
+    name: readonly
+    rules:
+      - verbs:
+          - get
+          - list
+        resources:
+          - checks
+          - entities
+'''
+
+RETURN = '''
+object:
+    description: object representing Sensu cluster role
+    returned: success
+    type: dict
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    arguments, errors, utils, role_utils
+)
+
+
+def validate_module_params(module):
+    params = module.params
+    if params['state'] == 'present':
+        if not params['rules']:
+            module.fail_json(
+                msg='state is present but all of the following are missing: rules'
+            )
+
+
+def main():
+    module = AnsibleModule(
+        supports_check_mode=True,
+        argument_spec=dict(
+            arguments.get_spec("auth", "name", "state"),
+            rules=dict(
+                type="list",
+                elements="dict",
+                options=dict(
+                    verbs=dict(
+                        required=True,
+                        type="list",
+                        choices=["get", "list", "create", "update", "delete"],
+                    ),
+                    resources=dict(
+                        required=True,
+                        type="list",
+                    ),
+                    resource_names=dict(
+                        type="list",
+                    ),
+                )
+            )
+        )
+    )
+
+    validate_module_params(module)
+    module.params['auth']['namespace'] = None  # Making sure we are not fallbacking to default
+    client = arguments.get_sensu_client(module.params["auth"])
+    path = "/clusterroles/{0}".format(module.params["name"])
+    payload = arguments.get_mutation_payload(
+        module.params, "rules"
+    )
+
+    try:
+        changed, cluster_role = utils.sync(
+            module.params['state'], client, path,
+            payload, module.check_mode, role_utils.do_roles_differ
+        )
+        module.exit_json(changed=changed, object=cluster_role)
+    except errors.Error as e:
+        module.fail_json(msg=str(e))
+
+
+if __name__ == '__main__':
+    main()
--- a/tests/integration/modules/molecule/cluster_role/playbook.yml
+++ b/tests/integration/modules/molecule/cluster_role/playbook.yml
@@ -1,0 +1,208 @@
+---
+- name: Converge
+  collections:
+    - sensu.sensu_go
+  hosts: all
+  gather_facts: no
+  tasks:
+    - name: Create a cluster role with missing required parameters
+      cluster_role:
+        auth:
+          url: http://localhost:8080
+        name: test_cluster_role
+      ignore_errors: true
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - "result.msg == 'state is present but all of the following are missing: rules'"
+
+    - name: Create a cluster role with empty rules
+      cluster_role:
+        auth:
+          url: http://localhost:8080
+        name: test_cluster_role
+        rules: []
+      ignore_errors: true
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - "result.msg == 'state is present but all of the following are missing: rules'"
+
+    - name: Create a cluster role with invalid rule verbs
+      cluster_role:
+        auth:
+          url: http://localhost:8080
+        name: test_cluster_role
+        rules:
+          - verbs:
+              - list
+              - do_something
+            resources:
+              - entities
+      ignore_errors: true
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+
+    - name: Create a cluster role with minimal parameters
+      cluster_role:
+        auth:
+          url: http://localhost:8080
+        name: minimal_test_cluster_role
+        rules:
+          - verbs:
+              - get
+              - list
+            resources:
+              - entities
+      register: result
+
+    - assert:
+        that:
+          - result is changed
+          - result.object.metadata.name == 'minimal_test_cluster_role'
+          - result.object.rules | length == 1
+          - result.object.rules.0.verbs | length == 2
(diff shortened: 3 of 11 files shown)


tests/integration/modules/molecule/cluster_role/playbook.yml
```diff
@@ -0,0 +1,194 @@
...
+
+    - name: Check idempotence of cluster role creation with minimal parameters
+      cluster_role:
+        auth:
+          url: http://localhost:8080
+        name: minimal_test_cluster_role
+        rules:
+          - verbs:
+              - get
+              - list
```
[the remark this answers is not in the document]
aljazkosir: nice catch :D, tests are failing if verbs are swapped. Custom comparator it is.

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,133 @@
...
+  - sensu.sensu_go.name
+  - sensu.sensu_go.state
+notes:
+  - Parameter C(auth.namespace) is ignored in this module.
+options:
+  rules:
+    description:
+      - Rules that the cluster role applies.
+    type: list
+    required: yes
```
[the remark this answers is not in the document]
mancabizjak: Indeed, it's confusing that the sanity checker is happy with or without this line. We should remove this here and in the `role` module as well, as `rules` is only conditionally required.

tests/integration/modules/molecule/cluster_role/playbook.yml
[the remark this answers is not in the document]
mancabizjak: :+1:  Great catch indeed, this one slipped us through for `role`.

tests/unit/modules/test_cluster_role_info.py
```diff
@@ -0,0 +1,47 @@
...
+)
+from ansible_collections.sensu.sensu_go.plugins.modules import cluster_role_info
+
+from .common.utils import (
+    AnsibleExitJson, AnsibleFailJson, ModuleTestCase, set_module_args,
+)
+
+
+class TestClusterRoleInfo(ModuleTestCase):
+    def test_get_all_roles(self, mocker):
```
mancabizjak: I would rename the method to `test_get_all_cluster_roles`, just to be consistent with naming in `test_cluster_role.py`.

tests/unit/modules/test_cluster_role_info.py
```diff
@@ -0,0 +1,47 @@
...
+        set_module_args()
+
+        with pytest.raises(AnsibleExitJson) as context:
+            cluster_role_info.main()
+
+        _client, path = get_mock.call_args[0]
+        assert path == "/clusterroles"
+        assert context.value.args[0]["objects"] == [1, 2, 3]
+
+    def test_get_single_role(self, mocker):
```
mancabizjak: Same as above, I suggest renaming to `test_get_single_cluster_role` for consistency.

aljazkosir:
I fixed the issues, I did some changes regarding the custom comparator for rules. I created a new file `rule_utils˙ as we talked about in other PR and also backed it with some tests. I would love some feedback on the suggested changes :).

tests/unit/modules/test_cluter_role.py
```diff
@@ -0,0 +1,122 @@
+from __future__ import absolute_import, division, print_function
```
mancabizjak: `test_cluter_role.py` -> `test_cluster_role.py` :scream: oops :grinning:

tests/unit/module_utils/test_role_utils.py
```diff
@@ -0,0 +1,145 @@
...
+            [{'a': [1, 3]}]
+        ) is True
+
+    def test_rules_more_rules_in_current(self):
+        assert role_utils._do_rules_differ(
+            [{'a': [2, 1]}, {'b': [3, 4]}],
+            [{'a': [1, 2]}]
+        ) is True
+
+    def test_rules_with_additional_keys_in_current(self):
```
mancabizjak: I may be wrong, but is this not basically the same test as `test_rules_more_rules_in_current`?

tests/unit/module_utils/test_role_utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class TestDoRulesDiffer:
+    def test_empty_values(self):
+        assert role_utils._do_rules_differ(
+            [{'a': []}],
+            [{'a': []}]
+        ) is False
+
+    def test_rules_when_value_is_none(self):
```
mancabizjak: I find this test a bit confusing, what exactly did you want to test here? Maybe it's just the actual test data that confuses me - I would at least make the value of `a` the same in current and desired rule set (or remove `a` entirely, if we want to test the behavior for `None` case in isolation)?

plugins/module_utils/role_utils.py
```diff
@@ -0,0 +1,43 @@
...
+
+def do_differ(current, desired):
+    if current is None:
+        return True
+
+    for key, value in desired.items():
+        if key == 'rules':
+            if _do_rules_differ(current['rules'], desired['rules']):
+                return True
+        elif value != current.get(key):
```
mancabizjak: Although I really like how nice and readable L38 is, we already have `desired['rules']` stored in `value` and we could also potentially store `current['rules']` (which is exactly `current.get(key)` from L40) to `current_value`. 

Something like:
```python
for key, value in desired.items():
    current_value = current.get(key)
    if key == 'rules':
        if _do_rules_differ(current_value, value):
            return True
    elif value != current_value:
           return True
```
I don't mind if you leave it as is though, I will have to slightly modify this part myself in #83 :slightly_smiling_face:.

tests/unit/module_utils/test_role_utils.py
```diff
@@ -0,0 +1,145 @@
...
+                'resources': ['hooks', 'assets'],
+                'resource_names': ['check-cpu']
+            }, {
+                'verbs': ['get', 'list'],
+                'resources': ['entities', 'checks']
+            }]
+        }
+        assert role_utils.do_differ(current, desired) is True
+
+    def test_rule_exists_but_with_additional_rules(self):
```
mancabizjak: I think you meant `test_role_exists_but_with_additional_rules`

tests/unit/module_utils/test_role_utils.py
aljazkosir: heh, you are right :sweat_smile:

tests/unit/module_utils/test_role_utils.py
aljazkosir: Yeah, I wanted to test the None cases, so `a` should not really matter in that case.

Maybe it would've be better if I used the actuall keys and values used in real case data instead of `a` and `b` and numbers.

plugins/module_utils/role_utils.py
aljazkosir: Heh, that condition was not in the loop at first, that's why it looks like this :sweat_smile:. I moved it there few seconds before the push. I'll change it, it looks better anyway :+1:

plugins/module_utils/role_utils.py
```diff
@@ -0,0 +1,43 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+def _sort_rules(rules):
+    for rule in rules:
+        for key in rule:
+            if rule[key]:
+                rule[key].sort()
+    return rules
```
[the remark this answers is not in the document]
aljazkosir: @tadeboro I tried with the suggested approach but I had some problems. If I used "one line loop" I was getting generators inside set, which made it impossible to compare. I solved it by implementing a normal for loop that adds a rule one at a time.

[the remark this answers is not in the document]
tadeboro: Generators inside sets? That sounds like a misplaced `)`. I would expect troubles with sets inside a tuple that is then placed in a set itself because the whole tuple becomes unhashable.

This is a bit more polished version of the code that I had in mind:

```
{
    (
        frozenset(r["verbs"] or []),
        frozenset(r["resources"] or []),
        frozenset(r["resource_names"] or [])
    ) for r in rules
}
```

Can you please check again what went wrong? Thanks.

plugins/module_utils/role_utils.py
```diff
@@ -42,3 +42,40 @@ def _do_subjects_differ(a, b):
     sorted_a = sorted(a, key=lambda x: (x['type'], x['name']))
     sorted_b = sorted(b, key=lambda x: (x['type'], x['name']))
     return sorted_a != sorted_b
+
+
+def _rule_sets(rules):
+    rule_set = set()
+    for rule in rules:
+        verbs = rule.get('verbs', []) or []
```
tadeboro: Is there a reason for that trailing `or []`? I thnk `get` should have us covered here.

plugins/module_utils/role_utils.py
```diff
@@ -42,3 +42,40 @@ def _do_subjects_differ(a, b):
     sorted_a = sorted(a, key=lambda x: (x['type'], x['name']))
     sorted_b = sorted(b, key=lambda x: (x['type'], x['name']))
     return sorted_a != sorted_b
+
+
+def _rule_sets(rules):
```
tadeboro: Maybe we should keep this name singular since we always return a single set?

plugins/module_utils/role_utils.py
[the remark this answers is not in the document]
aljazkosir: Hmm I see, one too many `()` :sweat_smile:.  Wouldn't be better to use `r.get('verbs', []) or []` for cases that key is completely missing?

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,131 @@
...
+  - For more information, refer to the Sensu documentation at
+    U(https://docs.sensu.io/sensu-go/latest/reference/rbac/)
+extends_documentation_fragment:
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.name
+  - sensu.sensu_go.state
+notes:
+  - Parameter C(auth.namespace) is ignored in this module.
+options:
+  rules:
```
tadeboro: Does sensu API accepts `rules: []`? And does ansible accept that thing as valid? This is getting a bit tricky ;)

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,131 @@
...
+    description:
+      - Rules that the cluster role applies.
+    type: list
+    suboptions:
+      verbs:
```
tadeboro: Same question as with rules: are empty lists OK?

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,131 @@
...
+        description:
+          - Permissions to be applied by the rule.
+        type: list
+        required: yes
+        choices: [get, list, create, update, delete]
+      resources:
```
tadeboro: Same question as with rules: are empty lists OK?

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,131 @@
...
+        description:
+          - Types of resources the rule has permission to access.
+        type: list
+        required: yes
+      resource_names:
```
tadeboro: Same question as with rules: are empty lists OK?

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,131 @@
...
+                    ),
+                    resource_names=dict(
+                        type="list",
+                    ),
+                )
+            )
+        )
+    )
+
+    module.params['auth']['namespace'] = None  # Making sure we are not fallbacking to default
```
tadeboro: That comment is the best!

plugins/module_utils/role_utils.py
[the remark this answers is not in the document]
tadeboro: Ugh, yes,  forgot that arguments are not always ansible-supplied and thus some keys can be missing. Great catch. And now you can ignore my comment in review about this ;)

plugins/modules/cluster_role.py
aljazkosir: Not valid on sensu API, but Ansible treats it as valid, even though `argument_spec` says that `verbs` and `resources` are required. If we set `required=True` on `rules` problems goes away... but `absent` state requires `rules` too then.
I think creating a custom validator function would be the best thing to do right here?

Also while I was looking into this I found out that a valid `verb` is also `['*']` which allows all permissions.

plugins/modules/cluster_role.py
aljazkosir: Sensu API seems to be OK with empty list, and Ansible too..., but to me it just doesn't make sense. This is a valid rule:
```
{'verbs': [], 'resources': [], 'resource_name': []}
```
Object saved on the API looks lie this:
```
{
    "metadata": {
        "name": "name"
    },
    "rules": [
        {
            "resource_names": null,
            "resources": null,
            "verbs": null
        }
    ]
}
```
Should we check for this kind of stuff in a custom validator too? Atleast for `verbs` and `resources` since they required on the API.
This applies to all three keys.

mancabizjak: I also noticed this behavior when implementing the `role` module. At that time, I didn't think about implementing custom validation to check whether `verbs` and `resources` are empty lists precisely because of what you reported, i.e. because the API seems to allow this case.

Nonetheless I am in favor of adding additional validations here and in `role` despite the API not breaking because of such cases (even when the docs state otherwise).

tadeboro: I asked upstream devs about this. I would say that for now, we keep things as they are (`[]` is a valid value for verbs, resources and resource names) and change them once we hear back from the sensu gals and guys. Does this sound OK with you, @mancabizjak and @aljazkosir?

plugins/modules/cluster_role.py
tadeboro: I also asked sensu devs about the semantics of the `*` value. Once we hear what they have to say about this, we will act appropriately, but for now, I would say that we just add it as a valid option and be done with it.

tadeboro: As for the roles being required, I vote for adding a custom validation into the module. This way we can at least produce a sensible error message while sensu API will probably just return a numeric error.

aljazkosir: Yeah, I'm greeted with classic 400 Bad request :grin:

plugins/modules/cluster_role.py
```diff
@@ -0,0 +1,137 @@
...
+)
+
+
+def validate_module_params(module):
+    params = module.params
+    if params['state'] == 'present':
+        if not params['rules']:
+            module.fail_json(
+                msg='state is present but all of the following are missing: rules'
+            )
```
tadeboro: I there a reason why this is not implemented using `required_if`?

aljazkosir: `required_if` only checks if value is `None`, here we also check if `rules` is empty array.

tests/integration/modules/molecule/cluster_role/playbook.yml
```diff
@@ -0,0 +1,208 @@
...
+
+    - name: Fetch all cluster roles
+      cluster_role_info:
+        auth:
+          url: http://localhost:8080
+      register: result
+
+    - assert:
+        that:
+          - result.objects | length == 8 # There are 6 default cluster roles
```
tadeboro: This will save us a ton of debugging once the Sensu changes defaul settings.

tadeboro:
@mancabizjak Once you clear this PR, can you please merge it? Thank you.

PR #89, opened 2019-11-06
Parallelize tests

tadeboro:
Changes in this commit heavily refactor the test suite layout in order to simplify it and make it parallelizable. Details about the changes are included in the individual commit messages.


HACKING.rst
```diff
@@ -34,28 +34,40 @@ version by running::
...
+configured, and running. And that is it. If nothing broke, we are ready to
+start developing.
+
+
+Section with content that will be reworked into a development story
+-------------------------------------------------------------------
+
+TODO: Describe the process of adding a new module.
+
    (venv) $ ansible-test sanity --python 3.7
```
mancabizjak: With the latest changes, this can be reduced to `$ make sanity`. 

EDIT: From the name of the section, I see that this content is subject to further polishing in the near future, but I think it makes sense to modify this here and now, since we now have more powerful `Makefile`s.

HACKING.rst
```diff
@@ -34,28 +34,40 @@ version by running::
...
    (venv) $ ansible-test units --python 3.7
```
mancabizjak: Similarly, can be reduced to `$ make units`

HACKING.rst
```diff
@@ -34,28 +34,40 @@ version by running::
...
    (venv) $ export ANSIBLE_COLLECTIONS_PATHS=$(pwd)/../../..
-   (venv) $ cd tests/integration/modules
-   (venv) $ molecule --base-config molecule/shared/base.yml test --all
+   (venv) $ cd tests/integration
+   (venv) $ molecule --base-config base.yml test -s module_asset
 
 .. note::
```
mancabizjak: If we update L59 and L60 so that they use `make` with new targets, this warning is not needed anymore as the Makefile takes care of determining python version for us.

tests/integration/Makefile
```diff
@@ -0,0 +1,17 @@
...
+.PHONY: all
+all: $(scenarios)
+
+.PHONY: modules
+modules: $(modules)
+
+.PHONY: roles
+roles: $(roles)
+
+.PHONY: $(modules) $(roles)
```
mancabizjak: I wasn't aware this is possible (and now that I'm looking at it - that it's actually pretty simple to achieve) - awesome! I always found myself temporarily hard-coding the module name somewhere...

HACKING.rst
```diff
@@ -34,28 +34,40 @@ version by running::
...
+
+TODO: Describe the process of adding a new module.
+
    (venv) $ ansible-test sanity --python 3.7
    (venv) $ ansible-test units --python 3.7
    (venv) $ export ANSIBLE_COLLECTIONS_PATHS=$(pwd)/../../..
-   (venv) $ cd tests/integration/modules
-   (venv) $ molecule --base-config molecule/shared/base.yml test --all
+   (venv) $ cd tests/integration
+   (venv) $ molecule --base-config base.yml test -s module_asset
```
mancabizjak: Alternatively: `$ make molecule/module_asset`

Makefile
```diff
@@ -17,12 +17,38 @@ endif
 # Make sure ansible can find our collection without having to install it.
 export ANSIBLE_COLLECTIONS_PATHS := $(realpath $(CURDIR)/../../..)
 
+python_version := $(shell \
```
mancabizjak: :+1:

HACKING.rst
```diff
@@ -113,3 +113,32 @@ Once we have our playbook ready, we can test our scenario by running::
...
+
+
+Continuous integration
+----------------------
+
+We use CircleCI to run our test suite on each push and pull request. Our CI
+pipeline is composed of two stages: fast one and slow one.
+
+The fast stage is composed of sanity and integration tests. Once those tests
+are done executing and passing, we start execution the slow stage that runs
```
aljazkosir:
```suggestion
are done executing and passing, we start executing the slow stage that runs
```

HACKING.rst
```diff
@@ -113,3 +113,32 @@ Once we have our playbook ready, we can test our scenario by running::
...
+``tests/integration/scenarios.times`` file. We can do this using a helper
+script ``tests/integration/time_scenarios.sh``. Note that it is not really
+required to run this script after each newly added integration test. It
+suffices to add it to the list manually and eyeball the testing time from
+other similar tests.
+
+When we execute the tests on the CI servers, we use those measurements to
+split the tests into partitions using the *longest processing time first*
+algorithm. Implementation is stored in the `tests/integration/partition.py``
+program. Nota that this program needs to be completely deterministic or some
```
aljazkosir:
```suggestion
program. Note that this program needs to be completely deterministic or some
```

HACKING.rst
tadeboro: Would it be OK with you if we keep this in for now? I promise I will move this into a more appropriate document under the `docs` directory as soon as possible.

Why would I like to keep this? While I really like the wrappers, I also feel that it is still immensely valuable to know how things work under the hood. Once things start to go south, knowing a thing or two about the internal workings can be a lifesaver. And yes, I know I can always dig this out of the git's history, I would instead move this content in a single commit.

mancabizjak: Sure, I understand the concern and can't say I disagree, so :+1: . I brought this up more for the sake of consistency with the first part of the updated `HACKING.rst` document which now includes instructions with `make`. 

Still, once we follow up with the dedicated document under `docs`, let's not forget to at least mention the possibility of using convenience wrappers as an alternative to running commands stated here :wink:

tadeboro: What I had in mind is to keep convenience wrappers in a quickstart and other casual documents that deal with tests on a higher level. Low level commands are destined for a document about trouble shooting the testing process.

tadeboro:
I think I addressed all of the comments in this PR.

I also introduced a last-minute change into our Molecule configuration. Because `flake8` and `ansible-test sanity` are both checking stylistic aspects of the code in an incompatible way, I deactivated the `flake8`. I decided to deactivate it because it was only checking the files Molecule knows about while sanity check covers all of the files in our repository.

mancabizjak:
Thank you, looks great :+1: !

PR #91, opened 2019-11-07
Custom comparator and params validation check for the role module

mancabizjak:
This PR includes a couple of fixes and minor updates in the `role` module, to bring it in sync with `cluster_role`:
* it adds a validation check to ensure `rules` parameter is provided and non-empty, same as in `cluster_role`*;
* it makes use of custom comparator provided by `role_utils`, to allow determination of roles' equality when `verbs`, `resources` or `resource_names` are equivalent but ordered differently.
* iz pluralizes resource names in integration tests so that they are aligned with valid Sensu resource types.

*Just thinking, would it make sense to move this validation to `role_utils` at some point?

### DIFF
--- a/plugins/modules/role.py
+++ b/plugins/modules/role.py
@@ -78,15 +78,6 @@ from ansible_collections.sensu.sensu_go.plugins.module_utils import (
 )
 
 
-def validate_module_params(module):
-    params = module.params
-    if params['state'] == 'present':
-        if not params['rules']:
-            module.fail_json(
-                msg='state is present but all of the following are missing: rules'
-            )
-
-
 def main():
     module = AnsibleModule(
         supports_check_mode=True,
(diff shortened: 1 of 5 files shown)


plugins/modules/role.py
```diff
@@ -75,16 +74,21 @@
...
 )
 
 
+def validate_module_params(module):
+    params = module.params
+    if params['state'] == 'present':
+        if not params['rules']:
+            module.fail_json(
+                msg='state is present but all of the following are missing: rules'
+            )
```
tadeboro: As you have already said, it would definitely make sense to move this check into common utility module. For example, we could add a method like this:

```
def validate_module_params(params):
    if params['state'] == 'present':
        if not params['rules']:
            return 'state is present but all of the following are missing: rules'
    return None
```

And then call it like

```
if msg := role_utils.validate_module_params(module.params):
    module.fail_json(msg=msg)
```

This way we can keep the utility functions pure and at least one developer happy ;)

mancabizjak: Done, albeit without the `:=` operator

mancabizjak:
To reduce overhead associated with the proposed change, I pushed another commit that extracts the validation function to the shared `role_utils`.

PR #96, opened 2019-11-12
Update tessen module

aljazkosir:
With this PR we update a tessen module to the new structure. We also needed to update `utils.put` function, since tessen API is returning code `200` instead of `201`, it makes sense since we are not really creating an object but only updating it's `opt_out` value.

Closes https://github.com/sensu/sensu-go-ansible/issues/85


tests/unit/modules/test_tessen.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class TestSync:
+    def test_remote_none(self, mocker):
+        client = mocker.Mock()
+        client.get.side_effect = (
+            response.Response(404, ''),
+            response.Response(200, '{"opt_out": "true"}'),
+        )
+        client.put.return_value = response.Response(201, "")
```
tadeboro: Is this a valid scenario? It feels a bit strange for API that deals with a static resource to return 404. If the API docs do not list the 404 as a valid return status, I would remove this test and add warn users about the failure. This probably means that we will need to use the Sensu client directly instead of relying on the helpers from `utils` module.

plugins/module_utils/utils.py
```diff
@@ -69,7 +69,7 @@ def delete(client, path):
 
 def put(client, path, payload):
     resp = client.put(path, payload)
-    if resp.status != 201:
+    if resp.status not in (200, 201):
```
tadeboro: If tessen is the only API endpoint that returns this status code, I would suggest that we leave the helper method alone and use the Sensu API client directly in the module. Weakening the method for the sake of single user feels a bit wrong.

tests/integration/molecule/module_tessen/playbook.yml
```diff
@@ -0,0 +1,53 @@
...
+          url: http://localhost:8080
+      ignore_errors: true
+      register: result
+
+    - assert:
+        that:
+          - result is failed
+          - "result.msg == 'missing required arguments: state'"
+
+    - name: Disable tessen
```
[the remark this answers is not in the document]
aljazkosir: I added a comment here, and a similar note to the module docs.

plugins/module_utils/utils.py
aljazkosir: Good point, I was thinking more in the direction that both 200 and 201 are valid codes for `HTTP PUT` (according to the RFC specification :)), that is why I put it there. However, as you pointed out, in this API that is mostly not the case.

PR #97, opened 2019-11-12
Rewrite README file in markdown

tadeboro:
One of the Ansible certification requirements mandates that we use
markdown as a documentation format. This is why we rewrote the README
document in markdown.

But because we would still like to host our more comprehensive
documentation somewhere on the web, we also copied the content into
our root Sphinx document. It is unfortunate that we have to have this
duplication in the codebase, but there is not much we can do about it.


README.md
```diff
@@ -0,0 +1,65 @@
...
+                 (Using Ansible Collections)
+   [sensu]: https://docs.sensu.io/sensu-go/latest
+            (Sensu Go documentation)
+
+
+## Installation
+
+If we would like to install the Sensu Go collection, we need to have Ansible
+2.9 or later installed. Once we have taken care of this, we can install the
+Sensu Go collection from [Ansible Galaxy][galaxy] by running:
```
mancabizjak: Oops, the link appears to be broken

tadeboro:
Link to relevant documentation: https://docs.ansible.com/ansible/latest/dev_guide/collections_galaxy_meta.html

README.md
```diff
@@ -0,0 +1,65 @@
...
+
+    $ ansible-galaxy collection install sensu.sensu_go
+
+   [galaxy]: galaxy.ansible.com (Ansible Galaxy)
+
+
+## Usage
+
+You can use the namespaced modules:
```
mancabizjak: These 2 ways (namespaced vs. bare names) of referencing the collection from playbooks are valid for roles too, right? If yes, then I think it wouldn't hurt to mention that this applies to roles too, just in case :)

README.md
```diff
@@ -0,0 +1,65 @@
+# Sensu Go Ansible Collection
+
+Sensu Go [Ansible Collection][collection] contains:
+
+* Ansible role to install [Sensu Go][sensu] on your nodes.
```
mancabizjak: Strictly speaking: roles, not role, right?  This sentence makes me think we're only referring to the role for installing agents, but there's the backend role as well.

docs/source/index.rst
```diff
@@ -1,4 +1,72 @@
...
+     tasks:
+       - name: List Assets
+         sensu.sensu_go.asset_info:
+         register: assets
+
+       - name: List Filters
+         sensu.sensu_go.filter_info:
+         register: filters
+
+Or add the collection to the search path and use bare names::
```
matejart:
```suggestion
Or use the `collections` keyword to create the search path, then use bare names:
```

docs/source/index.rst
```diff
@@ -1,4 +1,72 @@
...
+
+   $ ansible-galaxy collection install sensu.sensu_go
+
+.. _Ansible Galaxy: galaxy.ansible.com
+
+
+Usage
+-----
+
+You can use the namespaced modules::
```
matejart:
```suggestion
We can use the namespaced modules::
```

README.md
tadeboro: I think we inherited this text from the original author and never updated it. Good catch.

tadeboro:
Thanks for reviews. I updated the documents and pushed the changes.

docs/source/index.rst
```diff
@@ -1,4 +1,77 @@
-.. include:: ../../README.rst
+Sensu Go Ansible Collection
+===========================
+
+Sensu Go `Ansible Collection`_ contains:
+
+* Ansible role to install `Sensu Go`_ on your nodes.
```
mancabizjak: Roles ;)

tadeboro: I already hate the duplication ;)

mancabizjak:
LGTM, thanks @tadeboro!

Issue #95, opened 2019-11-10
Improve config file pemissions


mkobel:
The files `/etc/sensu/agent.yml` (agent role, main task) and `/etc/sensu/backend.yml` (backend role, main task) should be readable for sensu only because they can contain credentials (e.g. username/password for API).

With the current default permissions, these files are public readable.

tadeboro:
I am closing this since we merged the fixes @matejart prepared.

matejart:
The roles now create the configuration files owned by `sensu:sensu` and set permissions to `0600`. The change will be effective in the upcoming update of the collection.

PR #99, opened 2019-11-12
Protected access to backend and agent configuration files

matejart:
To address #95, we added the group and user ownership of the `/etc/sensu/backend.yml` and `/etc/sensu/agent.yml` files. We also set it to only be accessible to the `sensu:sensu` user and group.

### DIFF
--- a/roles/agent/tasks/main.yml
+++ b/roles/agent/tasks/main.yml
@@ -9,6 +9,10 @@
   template:
     src: agent.yml.j2
     dest: /etc/sensu/agent.yml
+    # Keep this in sync with what the agent service is running as from packager
+    owner: sensu
+    group: sensu
+    mode: '0600'
   notify: Restart agent
   tags: [configure_agent]
 
--- a/roles/backend/tasks/main.yml
+++ b/roles/backend/tasks/main.yml
@@ -9,6 +9,10 @@
   template:
     src: backend.yml.j2
     dest: /etc/sensu/backend.yml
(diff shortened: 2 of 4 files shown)


roles/backend/tasks/main.yml
```diff
@@ -9,6 +9,9 @@
   template:
     src: backend.yml.j2
     dest: /etc/sensu/backend.yml
+    owner: &sensu_user sensu
+    group: &sensu_group sensu
```
tadeboro: Do we intend to use user and group values anywhere else? If no, then we can safely drop the anchors. Also, can we add a short comment about the hardcoded user and group? Something like *keep this in sync with the user an group that the backend service is running as*, just to give future us some hints.

roles/agent/tasks/main.yml
```diff
@@ -9,6 +9,9 @@
   template:
     src: agent.yml.j2
     dest: /etc/sensu/agent.yml
+    owner: &sensu_user sensu
+    group: &sensu_group sensu
```
tadeboro: Again, anchors and a comment.

roles/backend/tasks/main.yml
matejart: Actually, we will use them later, but that doesn't mean that we have to have them here. I'll just drop this for this particular commit, because the anchors will have moved up anyway.
I'll add a comment.

matejart:
Thank you, @tadeboro.  The circleci should be green in a few minutes.

PR #100, opened 2019-11-14
Update event module [1/2]

aljazkosir:
With this commit, we update the `event_info` module to the new structure. API supports filtering by `entity` and by `entity` and `check` together, but does not support filtering by check, if this is needed we can filter this locally. That's why I created a separate PR for this, so we can discuss this here. Module is currently not backed with extensive integration test suite, due to the lack of event module. 

I plan to add a new commit or PR which will update the event module as well.


plugins/modules/event_info.py
```diff
@@ -0,0 +1,95 @@
...
+        ),
+    )
+
+    client = arguments.get_sensu_client(module.params['auth'])
+    if module.params['check_name']:
+        path = '/events/{0}/{1}'.format(module.params['entity_name'], module.params['check_name'])
+    elif module.params['entity_name']:
+        path = '/events/{0}'.format(module.params['entity_name'])
+    else:
+        path = '/events'
```
mancabizjak: Considering that the `/events` part of the path will always be present, I suggest removing `else` and move this line before L77. Then, if necessarry (i.e. depending on the provided params), you can append the appropriate trailing part of the path.

This way we can reduce `if-elif-else` to `if-elif` :slightly_smiling_face:

plugins/modules/event_info.py
```diff
@@ -0,0 +1,95 @@
...
+'''
+
+EXAMPLES = '''
+- name: List Sensu events
+  event_info:
+  register: result
+
+- name: List Sensu events for api.example.com
+  event_info:
+    entity: api.example.com
```
mancabizjak:
```suggestion
    entity_name: api.example.com
```
But to be honest, I think I prefer `entity` and `check` over `entity_name` and `check_name` :grinning: Whatever we decide to go for, let's keep in mind to make example and docs/argument specs consistent :)

plugins/modules/event_info.py
```diff
@@ -0,0 +1,95 @@
...
+
+def main():
+    required_by = {'check_name': ['entity_name']}
+    module = AnsibleModule(
+        supports_check_mode=True,
+        required_by=required_by,
+        argument_spec=dict(
+            arguments.get_spec("auth"),
+            check_name=dict(),
+            entity_name=dict(),
```
mancabizjak: Despite reading the docs at the beginning of this file first,  the example left me a bit confused. I had to stop and think for a second about what `entity_name` and `check_name` are. Somehow I felt like we're actually trying to configure something (not filters!), even though I know `*_info` modules never change the state of resources on the backend.

If I were the user of the module, I think I'd prefer writing a task like this:
```yaml
- name: List Sensu events for api.example.com
  event_info:
    filters:
     - entity: api.example.com  # or entity_name
     - check: some-check        # or check_name
  register: result
```

Grouping supported filters under optional `filters` parameter seems a bit more intuitive to me, and I feel like it expresses our intentions better. But maybe not all end-users are as confused as me ;)

What do you guys think? @tadeboro

tests/unit/modules/test_event_info.py
```diff
@@ -0,0 +1,64 @@
...
+            check_name='sample-check'
+        )
+
+        with pytest.raises(AnsibleExitJson) as context:
+            event_info.main()
+
+        _client, path = get_mock.call_args[0]
+        assert path == '/events/sample-entity/sample-check'
+        assert context.value.args[0]['objects'] == [4]
+
```
mancabizjak: I think we're missing a test case to check that the module fails when when `check_name` is provided but `entity_name` isn't. Just to be sure validation of module params kicks in the way it's supposed to :)

EDIT: ignore this comment if we decide to support local filtering based on `check_name` alone.

plugins/modules/event_info.py
tadeboro: On the one hand, I kind of like the `filters` approach. It is explicit about what we are doing. And we can quickly transform it into something that can retrieve events from different sources. I would add one little change to the proposed API. Because the `check` parameter is dependent on `entity` being present, I would group them like this:

    - event_info:
        filters:
         - entity: first_enity
           check: first_check
         - entity: second_entity
         - entity: third_enity
           check: third_check
      register: result

In a way, the `entity` and `check` pair is a single filter.

But on the other hand, if we stick with a top-level `entity` and `check` parameters, we can retain the symmetry that all other module pairs have. For example, we create an asset like this:

    - asset:
        name: my_asset
        # other params here

If we want to retrieve the information about the asset we created, we add an `_info` suffix and drop everything except the name:

    - asset_info:
        name: my_asset

Now, the event module's API will look something like this (@aljazkosir, please correct me if I am wrong):

    - event:
        entity: my_entity
        check: my_check
        # other params here

`entity` and `check` parameter pair has a very similar function as the `name` parameter in other modules. It feels natural that we query about the events in roughly the same way:

    - event_info:
        entity: my_entity
        check: my_check

Now, I know that events are a bit special in that there can be more than one instance that matches the `entity` and `check` parameters, but the API still feels kind of right.

Again, I am fine with any of the approaches here. Maybe I am a bit biased towards the second version (where the `entity` and `check` are not nested under `filters`) because it feels it should be less code to implement. And less code is good ;) But if you both vote for the other approach, we will go with that.

plugins/modules/event_info.py
tadeboro: Or we can go almost completely `if`-less with

    path_components = '/events', module.params['entity'], module.params['check']
    path = '/'.join(c for c in path_components if c)

But this is just me golfing, so feel free to ignore my code ;)

tests/unit/modules/test_event_info.py
tadeboro: I do not think Sensu Go API allows us to filter by check-only, so that test is indeed needed.

plugins/modules/event_info.py
mancabizjak: That's way cooler! :+1:

tests/unit/modules/test_event_info.py
aljazkosir: Ops, I added a similar test in integration tests but forgot about the unit tests.

plugins/modules/event_info.py
aljazkosir: I couldn't decide between the two, it seems like I forgot to change all the occurrences :grin:. Anyway, I removed the `_name` prefix.

tests/integration/molecule/module_event/playbook.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: Converge
+  collections:
+    - sensu.sensu_go
+  hosts: all
+  gather_facts: no
+  tasks:
+    - name: Event info without entity_name
```
mancabizjak: Nitpick: without entity :)

tadeboro:
Not sure why `apt` command failed (feels like network hiccup). I restarted the tests to confirm.

PR #101, opened 2019-11-14
Update cluster_role_binding and cluster_role_binding_info modules

mancabizjak:
This commit applies the new module structure to `cluster_role_binding` and `cluster_role_binding_info` modules. Cluster role bindings allow assignment of cluster roles to users and groups.

To avoid duplication of module params validation code (which is exactly the same for both `role_binding` and `cluster_role_binding` modules), I added a slightly modified version of the validator previously in `role_binding.py` and moved it to common `utils/role_utils`, where it can be used by both modules.

### DIFF
--- a/plugins/module_utils/role_utils.py
+++ b/plugins/module_utils/role_utils.py
@@ -14,6 +14,12 @@ def validate_module_params(params):
     return None
 
 
+def validate_binding_module_params(params):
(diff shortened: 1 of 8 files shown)


gberginc:
@aljazkosir but what if that was just a test? I am always suspicious when @tadeboro says something like this :-)

PR #102, opened 2019-11-15
Add support for cluster roles in the role_binding module

mancabizjak:
From [Sensu Go Role Bindings reference](https://docs.sensu.io/sensu-go/5.14/reference/rbac/#role-bindings-and-cluster-role-bindings):  
> A role binding assigns **a role or cluster role** to users and groups within a namespace.

ATM the `role_binding` module does not allow applying cluster roles in the context of the current namespace, as the `role` parameter expects a "normal" role that acts on namespaced resources (i.e. not a cluster role). In this PR we update the module so that we can specify `cluster_role` as well. The new parameter is mutually exclusive with `role`.

### DIFF
--- a/plugins/modules/role_binding.py
+++ b/plugins/modules/role_binding.py
@@ -31,17 +31,25 @@ options:
   role:
     description:
       - Name of the role
+      - This parameter is mutually exclusive with C(cluster_role).
+    type: str
+  cluster_role:
+    description:
+      - Name of the cluster role. Note that the resulting role
+        binding grants the cluster role to the provided users and
+        groups in the context of C(auth.namespace) only.
+      - This parameter is mutually exclusive with C(role).
     type: str
   users:
     description:
-      - List of users to bind to the role
-      - Note that at least one of 'users' and 'groups' must be
+      - List of users to bind to the role or cluster role
+      - Note that at least one of C(users) and C(groups) must be
         specified when creating a role binding.
     type: list
   groups:
     description:
-      - List of groups to bind to the role
-      - Note that at least one of 'users' and 'groups' must be
+      - List of groups to bind to the role or cluster role
+      - Note that at least one of C(users) and C(groups) must be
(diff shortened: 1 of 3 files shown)


plugins/modules/role_binding.py
```diff
@@ -72,24 +88,35 @@
 )
 
 
+def infer_role(params):
```
tadeboro: Can we add two small test cases for this function? I know we already test it indirectly when performing other checks, but I would like to see those tests there just for the sake of completeness. Something like this should do:

```
@pytest.mark.parametrize("params,result", [
    (
        dict(role="my_role"),
        ("Role", "my_role"),
    ),
    (
        dict(cluster_role="my_cluster_role"),
        ("ClusterRole", "my_cluster_role"),
    ),
])
def test_infer_role(self, params, result):
    assert result == infer_role(params)
```

tests/unit/modules/test_role_binding.py
```diff
@@ -137,6 +170,18 @@ def test_failure(self, mocker):
         with pytest.raises(AnsibleFailJson):
             role_binding.main()
 
+    def test_failure_role_and_cluster_role(self, mocker):
+        sync_mock = mocker.patch.object(utils, 'sync')
+        sync_mock.side_effect = Exception("Validation should fail but didn't")
```
tadeboro: This is the stuff that makes it possible for a developer to sleep at night ;)

plugins/modules/role_binding.py
mancabizjak: Done. In the test cases, I'm also explicitly setting the other parameter to `None`, as this is the default case when we specify a parameter in module's argument spec and don't set it:
```
dict(cluster_role="a", role=None)
```

mancabizjak:
Thanks @tadeboro! I added the test to ensure `infer_role` behaves as expected.

Issue #79, opened 2019-10-24
Prepare event and event_info modules


tadeboro:
Initial work has been done in #54, but this needs a major simplification.

PR #104, opened 2019-11-19
Update event module

aljazkosir:
With this commit, we update the event module to the new structure. Module requires check and entity objects to exists on the remote beforehand. 

Closes https://github.com/sensu/sensu-go-ansible/issues/79


tadeboro:
I only did a preliminary review, and the module's API is all over the place.

I blame this one on the Sensu's upstream API, and I think that in this case, we should create a more sensible alternative. The API that I came up with after a few minutes of thinking about this problem looks like this:

```
- event:
    entity: my_entity_name
    check: my_check_name
    check_attributes:
      # Stuff from https://docs.sensu.io/sensu-go/5.14/reference/events/#check-attributes
    metric_attributes:
      # Stuff from https://docs.sensu.io/sensu-go/5.14/reference/events/#metric-attributes
```

It feels more natural (at least to me) to have event-specific things under separate parameters. And the upstream's documentation tends to agree with me since the links fit right in ;)

@mancabizjak @aljazkosir Do you have any suggestions around the API structure?

tadeboro:
As for the metric parameter, I think metric-only events are not persisted and thus cannot be retrieved using the API. Part of the official documentation that I based my opinion is [here](https://docs.sensu.io/sensu-go/5.14/reference/events/#metric-only-events), and the most relevant sentence seems to be:

> These events, since they do not contain checks, bypass the store, and are sent off to the event pipeline and corresponding event handlers.

So I would say we still include those parameters in the API and inform users about the specifics.

aljazkosir:
@tadeboro I agree with your suggestion. Even when I was writing the module it felt weird, because check needed a `name` parameter while `entity` didn't. But I went with it because I think that's how we decided few months ago. And yes `check_attributes` field fits right in. :grin:

tests/integration/molecule/module_event/playbook.yml
```diff
@@ -18,6 +18,187 @@
...
+        that:
+          - result.objects | length == 1
+
+    - name: Create second entity
+      entity:
+        auth:
+          url: http://localhost:8080
+        name: entity2
+        entity_class: proxy
+      register: entity
```
mancabizjak: I could be wrong, but I think we're not using the returned `entity` anywhere, so we can we can remove this?

tests/integration/molecule/module_event/playbook.yml
```diff
@@ -18,6 +18,187 @@
...
+      check:
+        auth:
+          url: http://localhost:8080
+        name: check2
+        command: /bin/false
+        subscriptions:
+          - checks
+          - also_checks
+        interval: 30
+      register: check
```
mancabizjak: Same as for `register: entity` - I think we can remove this.

plugins/modules/event.py
```diff
@@ -0,0 +1,269 @@
...
+    subscriptions:
+      - checks
+      - also_checks
+    interval: 30
+  register: check
+- name: Create event with minimal parameters
+  event:
+    auth:
+      url: http://localhost:8080
+    entity: awesome_entity
```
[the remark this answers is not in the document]
aljazkosir: Ah of course... At first, I did it like you suggested (with `{{ check.object.metadata.name }}`), but it seemed ugly. So I used simple names, but then I forgot to remove the `register`s.

tests/integration/molecule/module_event/playbook.yml
```diff
@@ -18,6 +18,187 @@
...
+              status: 0
+            - executed: 1552505493
+              status: 0
+          issued: 1552506034
+          last_ok: 1552506033
+          output: '10'
+          state: 'passing'
+          status: 'ok'
+          total_state_change: 0
+        event_attributes:
```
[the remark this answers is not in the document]
aljazkosir: How did I miss that :scream:. Even the tests failed.

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+- name: Create a simple check
+  check:
+    auth:
+      url: http://localhost:8080
+    name: awesome_check
+    command: echo "Hello world"
+    subscriptions:
+      - checks
+      - also_checks
+    interval: 30
```
tadeboro: I think we can skip this whole thing since we only need entity and check names in the current example. Is this a remnant from the good old times when our API was structured in such a way that we required "fully hydrated" objects in event module?

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+    if resp.status != 200:
+        raise errors.SyncError("Entity with name '{0}' does not exist on remote.".format(entity))
+    return resp.json
+
+
+def _update_payload_with_metric_attributes(payload, metric_attributes):
+    if metric_attributes:
+        payload['metrics'] = arguments.get_spec_payload(metric_attributes,
+                                                        *metric_attributes.keys())
+    return payload
```
tadeboro: This return is not really needed because we are mutating the `payload` dictionary in-place.

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+
+
+def _update_payload_with_check_attributes(payload, check_attributes):
+    if check_attributes:
```
tadeboro: We can get rid of one level of indentation by inverting the condition and exiting function early.

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+        if check_attributes['status']:
+            check_attributes['status'] = STATUS_MAP[check_attributes['status']]
+
+        filtered_attributes = arguments.get_spec_payload(check_attributes, *check_attributes.keys())
+        payload['check'].update(filtered_attributes)
+    return payload
```
tadeboro: Again, not really needed.

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+def _build_api_payload(client, params):
+    payload = arguments.get_spec_payload(params, 'timestamp')
+    payload['metadata'] = dict(
+        namespace=params['auth']['namespace']
+    )
+    payload['entity'] = _get_entity(client, params['entity'])
+    payload['check'] = _get_check(client, params['check'])
+
+    payload = _update_payload_with_check_attributes(payload, params['check_attributes'])
+    payload = _update_payload_with_metric_attributes(payload, params['metric_attributes'])
```
tadeboro: We can drop the assignments because both functions mutate payload in-place.

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+    return payload
+
+
+def sync(client, path, payload, check_mode):
```
tadeboro: The name of this function is a bit misleading. We are not actually syncing anything, we simply send the event. This is why I would suggest renaming this into `send` or something similar.

tests/unit/modules/test_event.py
```diff
@@ -0,0 +1,279 @@
...
+            check=dict(
+                metadata=dict(
+                    name='awesome_check',
+                    namespace='default'
+                )
+            )
+        )
+        assert check_mode is False
+
+    def test_all_event_parameters(self, mocker):
```
tadeboro: That is one monster of a test! Thanks for your perseverance, it is awesome!

plugins/modules/event.py
```diff
@@ -0,0 +1,296 @@
...
+        raise errors.SyncError("Check with name '{0}' does not exist on remote.".format(check))
+    return resp.json
+
+
+def _get_entity(client, entity):
+    entity_path = '/entities/{0}'.format(entity)
+    resp = client.get(entity_path)
+    if resp.status != 200:
+        raise errors.SyncError("Entity with name '{0}' does not exist on remote.".format(entity))
+    return resp.json
```
tadeboro: Since we mock those two functions later in the tests, maybe we should remove the leading `_` in the names? It feels a bit wrong to mock something that is marked as ~none of your business~ private ;) Besides; you already wrote the tests (which are not needed nor desired for private functions), so your gut feeling already guided you in the correct direction.

plugins/modules/event.py
aljazkosir: I left this as an example so end-user can see that `entity` and `check` must exist, but yeah, that's already mentioned in the docs.

plugins/modules/event.py
aljazkosir: Heh, thanks, I was going back and forth and couldn't decide between "private" and "public".

plugins/modules/event.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def _update_payload_with_metric_attributes(payload, metric_attributes):
+    if not metric_attributes:
+        return None
```
tadeboro: Just a tiny nitpick: if the function returns `void`, it should return `void` in all branches. Now, because python does not know how to return `void`, it returns `None` in such cases. This makes this line valid, but it would be considered more correct to just call `return` with no parameters, since this gives another indication that the return value of this function is not meant to be used.

plugins/modules/event.py
```diff
@@ -0,0 +1,279 @@
...
+
+    payload['metrics'] = arguments.get_spec_payload(metric_attributes, *metric_attributes.keys())
+
+
+def _update_payload_with_check_attributes(payload, check_attributes):
+    if not check_attributes:
+        return None
```
tadeboro: Same nitpick as in previous comment.

tests/unit/modules/test_event.py
```diff
@@ -0,0 +1,279 @@
...
+)
+from ansible_collections.sensu.sensu_go.plugins.modules import event
+
+from .common.utils import (
+    AnsibleExitJson, AnsibleFailJson, ModuleTestCase, set_module_args,
+)
+
+
+class TestGetObjects:
+    def testget_entity(self, mocker):
```
mancabizjak: Was this accidentally find-and-replaced when `_get_entity` became `get_entity`? ;) Same for both entity and both check tests :)

aljazkosir: heh whoops :blush:

PR #107, opened 2019-11-22
Update asset module API to support multiple-build assets

mancabizjak:
This PR modifies the API of the asset module to allow creation of multiple-build assets.

 From now on `url`, `sha512`, `filters` and `headers` are no longer top-level arguments of the module, as all of these must/can be provided for individual builds. We introduce the `builds` argument that holds a list of builds with the same argument specification as the prevous single-build specification of the asset.

While the Sensu Go backend API allows top-level `filters` and `headers` when creating an asset in the latest version (`5.15` at the time of writing), this seems to be for backwards-compatibility with the older versions. In the module, we drop support for these in favor of per-build `headers` and `filters` only.

### DIFF
--- a/plugins/modules/asset.py
+++ b/plugins/modules/asset.py
@@ -19,6 +19,7 @@ module: asset
 author:
   - Cameron Hurst (@wakemaster39)
   - Aljaz Kosir (@aljazkosir)
+  - Manca Bizjak (@mancabizjak)
   - Miha Plesko (@miha-plesko)
   - Tadej Borovsak (@tadeboro)
 short_description: Manages Sensu assets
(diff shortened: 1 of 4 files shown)


plugins/modules/asset.py
```diff
@@ -90,9 +103,58 @@
...
+
+def _sorted_builds(builds):
+    build_tuples = [(
+        b.get('sha512'),
+        b.get('url'),
+        b.get('headers', {}),
+        frozenset(b.get('filters', []) or []),
+    ) for b in builds]
+
+    return sorted(build_tuples, key=lambda x: (x[0]))
```
tadeboro: Alternative implementation (using sets) that does not need sorting would be:

```
    return {(
        b.get('sha512'),
        b.get('url'),
        frozenset((b.get('headers', {}) or {}).items()),
        frozenset(b.get('filters', []) or []),
    ) for b in builds}
```

Now we can simply compare the results of this function using reqular `==` and `!=`.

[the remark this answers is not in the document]
mancabizjak: Ugh, I just blindly assumed that SHA512 would always be different. Thanks.

mancabizjak:
Thanks @tadeboro! I applied the suggested changes related to comparison of builds.

 I also added another unit test for `_do_builds_differ` that compares two equal builds with `headers`/`filters` explicitly set to `None`. Just to be sure we're really handling such cases as expected :slightly_smiling_face:

plugins/modules/asset.py
```diff
@@ -90,9 +103,54 @@
...
+def do_differ(current, desired):
+    if current is None:
+        return True
+
+    for key, value in desired.items():
+        current_value = current.get(key)
+        if key == 'builds':
+            if _do_builds_differ(current_value, value):
+                return True
+        elif value != current.get(key):
```
tadeboro: Nitpick: a bit more efficient version would be `elif value != current_value:` since we already fetched it earlier.

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
...
+                        "url": "http://def.com",
+                        "sha512": "abc",
+                        "filters": ["abc == def"],
+                    }
+                ]
+            },
+        ) is True
+
+
+class TestDoBuildsDiffer:
```
tadeboro: Would it be possible to merge tests in the `TestDoBuildsDiffer` class into the `TestDoDiffer` set of tests? There is a significant amount of overlap between those two sets of tests that we can safely remove. And then we can stop testing the private method ;)

See further comments for details.

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
...
+                "sha512": "a",
+                "url": "a",
+                "filters": ["a", "b", "c"],
+            }],
+            [{
+                "sha512": "a",
+                "url": "a",
+                "filters": ["c", "a", "b"],
+            }],
+        ) is False
```
tadeboro: We already cover those two scenarios with the `TestDoDiffer.test_equal` test case. We can safely drop them.

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
...
+                    "url": "a",
+                    "sha512": "a",
+                    "filters": ["b", "c", "a"],
+                    "headers": {
+                        "bar": "foo",
+                        "foo": "bar",
+                    },
+                },
+            ],
+        ) is False
```
tadeboro: This scenario is also almost covered. If we add a few more entries to the `headers` dictionary in the `TestDoDiffer.test_equal` case, we can drop this one.

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
...
+                "sha512": "a",
+                "url": "a",
+                "filters": None
+            }],
+            [{
+                "sha512": "a",
+                "url": "a",
+                "headers": None,
+            }],
+        ) is False
```
tadeboro: This test is great we should move it into the `TestDoDiffer` class.

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
...
+
+    def test_different_lengths(self):
+        assert asset._do_builds_differ(
+            [{"sha512": "a", "url": "a"}],
+            [
+                {"sha512": "a", "url": "a"},
+                {"sha512": "b", "url": "b"}
+            ],
+        ) is True
```
tadeboro: This scenario is covered by `TestDoDiffer.test_different`. And while we are at it, maybe we should rename that test to `test_different_length` to be a bit more informative with the name?

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
...
+                {
+                    "url": "b",
+                    "sha512": "b",
+                    "filters": ["a", "b", "c"],
+                    "headers": {
+                        "abc": "abc",
+                    },
+                },
+            ],
+        ) is True
```
tadeboro: We should move this test and add rename it to `test_different_build_content` to give us a hint as to where we should look for changes.

tests/unit/modules/test_asset.py
```diff
@@ -13,25 +14,242 @@
 )
 
 
+class TestDoDiffer:
```
tadeboro: Would it be possible to add another test that has equal builds but one of the other fields is different?

tests/unit/modules/test_asset.py
mancabizjak: I don't feel very strongly about whether one should or shouldn't test private methods, so I'm good with either keeping `TestDoBuildsDiffer` or merging it with `TestDoDiffer` (while removing what's duplicated between them).  

I do love consistency though, and I thought I should take the same approach as we did with the [test suite for `role_utils`](https://github.com/sensu/sensu-go-ansible/blob/master/tests/unit/module_utils/test_role_utils.py#L12), where we test both private `_do_subjects_differ` method and `do_role_bindings_differ` as the only consumer of the same private method. But I do see now, that there's not as much overlap there :)

tadeboro: As my original comment suggested, I am mostly concerned with the duplication since it makes any future changes even more painful.

As for the testing of private methods, this is just something that I try not to do since some linters (like pylint with default settings) yell at me ;) But in this case, pylint is tamed enough not to bark about this, so feel free to do whatever feels right to you.

Oh, and about that role utils tests: I must have missed that one ;) And now that I know about it, it does bother me a bit ;)

mancabizjak:
I removed `TestDoBuildsDiffer` and modified tests in `TestDoDiffer`. @tadeboro Please take another look at the updated test suite, I think it's more manageable now :)

PR #108, opened 2019-11-25
Add bonsai asset action plugin

tadeboro:
The main purpose of this pull request is to add the `bonsai_asset` action plugin. Unfortunately, it was quite a journey through the collection internals in order to get them ready for new functionality that we need in this module.

For more details, see commit messages.


tadeboro:
Just a quick heads up: I moved the URL path changes into separate PR (#110). Changes in that commit could be seen as a bug fix since we failed to create resources with valid names according to the Sensu backend.

tadeboro:
> As for the `bonsai_asset` in general, it's super useful! I'm sharing some thoughts here, reflecting a bit on how we could take things a bit further still.
> When reading module docs, I noticed the user is instructed to use regular `asset` module for deletion of assets. This is OK if he wants to delete the whole asset, but what if he wants to delete a specific version? Sensu Go backend can't handle this, right?

The Sensu backend knows nothing about versions. If we would like to know anything about that, we need to look for the various semi-standard annotations that all Bonsai assets are equipped with. We only use the version parameter to fetch the right asset definition from the Bonsai server and then we discard it.

For example, in the following scenario, the second task will replace the asset that was created by the first task:

```
- bonsai_asset:
    name: a/b
    version: 1.2.3

- bonsai_asset:
    name: a/b
    version: 1.2.4
```

> Similarly, it seems that we can't get asset info for a particular version of the asset from Bonsai. Will we need to add `bonsai_asset_info` module to retrieve such info? I think the integration test would certainly benefit from it, since it would be great if the test could show that 2 versions of the same asset co-exist on Bonsai, like this:

Currently, different versions cannot coexist on the backend. This limitation stems from the fact that we do not store version information anywhere within the name. We could definitely add this (by appending the version number to the name of the asset) but I opted not to go down that route. The main reason for my decision is compatibility with the `sensuctl asset add` command. When we add assets using `sensuctl asset add` command, the version is also discarded and only name is used.

There is also another way we can get different versions of an asset to coexist on the backend: by making it possible for the playbook author to rename the bonsai assets. This is the route `sensuctl` command took:

```
$ ./sensuctl asset add
adds an asset definition fetched from Bonsai

Usage:	sensuctl asset add [NAME] [flags]

Flags:
  -h, --help            help for add
  -r, --rename string   rename the asset to the provided string after fetching it from Bonsai

# More bla bla here
```

But now that I am thinking about it, maybe it would be nice to be able to say

    - bonsai_asset:
        bonsai_name: sensu/monitoring-plugins
        version: 1.2.3
        name: sensu-monitoring-plugins-1.2.3

or


    - bonsai_asset:
        name: sensu/monitoring-plugins
        version: 1.2.3
        rename: sensu-monitoring-plugins-1.2.3

mancabizjak:
> There is also another way we can get different versions of an asset to coexist on the backend: by making it possible for the playbook author to rename the bonsai assets. This is the route `sensuctl` command took

From the usability standpoint, I think we should support coexistence of multiple asset versions on the Sensu backend. Especially since usability and convenience are the motivating factors behind the `bonsai_asset` module (you know how bonsai asset definitions look like without it ;))

I prefer the second suggestion with `rename`:

> ```
> bonsai_asset:
>   name: sensu/monitoring-plugins
>   version: 1.2.3
>   rename: sensu-monitoring-plugins-1.2.3
> ```

as the first one seems a bit less clear to me.

I also like that with the proposed change we wouldn't need additional `bonsai_asset_info` module as our `asset_info` would work perfectly good - as long as we're clear that the created asset should be referenced under `rename`d name later on.

tadeboro:
It took me a while, but I managed to get the ready-for-review code into this PR.

tadeboro:
There is still one extra commit in this PR that I will remove as soon as #110 is merged.

tests/unit/action/test_bonsai_asset.py
```diff
@@ -0,0 +1,243 @@
...
+
+class TestValidate:
+    @pytest.mark.parametrize("name,args,required,typ", [
+        # Required values must match the selected type.
+        ("a", dict(a=3), True, int),
+        ("a", dict(a=3.3), True, float),
+        ("a", dict(a="b"), True, str),
+        ("a", dict(a=[]), True, list),
+        ("a", dict(a={}), True, dict),
+        # Optional values are not checked for type-correctness.
```
mancabizjak: I find this comment a bit confusing. Based on the comment, I was expecting to see cases like:
 ```
("a", dict(a="b"), False, int) #  optional parameter "a" is given and has wrong type
```
that would pass. But cases like the one above fail (as they should). I'm not sure what you meant here.

Perhaps we need another set of cases to ensure that we have optional parameters which are provided and of the correct type, like this:
```
("a", dict(a=1), False, int)
("a", dict(a=1.2), False, float)
("a", dict(a="b"), False, str)
# ...
```

File: tests/unit/action/test_bonsai_asset.py [hunk 2]
```diff
@@ -0,0 +1,243 @@
...
+        ("a", dict(), False, dict),
+    ])
+    def test_valid_values(self, name, args, required, typ):
+        bonsai_asset.validate(name, args, required, typ)
+
+    def test_missing_required(self):
+        with pytest.raises(errors.Error, match="required"):
+            bonsai_asset.validate("a", {}, True, str)
+
+    def test_invalid_type(self):
```
mancabizjak: Can we also add a test where optional parameter has invalid type?

tests/unit/action/test_bonsai_asset.py
```diff
@@ -0,0 +1,243 @@
...
+            name="my-asset",
+            state="present",
+            builds=[],
+        )
+
+    def test_fail_bonsai(self, mocker):
+        bonsai_params = mocker.patch.object(bonsai, "get_asset_parameters")
+        bonsai_params.side_effect = errors.BonsaiError("Bonsai bad")
+
+        with pytest.raises(errors.Error, match="Bonsai bad"):
```
mancabizjak: How about `with pytest.raises(errors.BonsaiError)`?

tests/unit/action/test_bonsai_asset.py
```diff
@@ -0,0 +1,243 @@
...
+            ))
+
+    def test_invalid_annotations(self):
+        with pytest.raises(errors.Error, match="annotations"):
+            bonsai_asset.ActionModule.validate_arguments(dict(
+                name="abc", version="1.2.3", annotations=1,
+            ))
+
+
+class TestBuildAssetArgs:
```
mancabizjak: Can we add a test to check that `auth` is in the result if provided as argument?

tests/unit/action/test_bonsai_asset.py
```diff
@@ -0,0 +1,243 @@
+from __future__ import absolute_import, division, print_function
```
mancabizjak: Nitpick: I suggest renaming the directory to `tests/unit/actions` (note the plural), as we already use plural with `modules` and `module_utils` at the same level.

plugins/action/bonsai_asset.py
```diff
@@ -0,0 +1,97 @@
...
+            asset_args = self.build_asset_args(self._task.args)
+            return merge_hash(
+                result,
+                self._execute_module(
+                    module_name="sensu.sensu_go.asset", module_args=asset_args,
+                    task_vars=task_vars, wrap_async=wrap_async,
+                ),
+            )
+        except errors.Error as e:
+            print("AKHDKASDKJGA")
```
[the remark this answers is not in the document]
tadeboro: Let us just pretend that this did not happen ;)

plugins/action/bonsai_asset.py
```diff
@@ -0,0 +1,97 @@
...
+    value = args.get(name)
+    if required and value is None:
+        raise errors.Error("{0} is required argument".format(name))
+    if value is not None and not isinstance(value, typ):
+        raise errors.Error("{0} should be {1}".format(name, typ))
+
+
+class ActionModule(ActionBase):
+
+    TRANSFERS_FILES = False
```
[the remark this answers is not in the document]
tadeboro: I initially started with a different base class and forgot to remove this I was when simplifying things. Great catch.

tests/integration/molecule/action_bonsai_asset/playbook.yml
```diff
@@ -0,0 +1,87 @@
+---
+- name: Converge
+  collections:
+    - sensu.sensu_go
+  hosts: all
+  gather_facts: no
+  tasks:
+    - name: Create a Bonsai asset
+      bonsai_asset: &idempotence
```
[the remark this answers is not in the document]
tadeboro: I did not think about this at all. Defaults are so deeply ingrained into my brains that I automatically left the parameter out. This is what I get for not copy-pasting the existing tests and modifying them afterwards ;) I will add `auth` parameter.

tests/unit/action/test_bonsai_asset.py
tadeboro: I initially named this folder `actions`, but then I changed my mind and decided to keep it consistent with the name in the `plugins` directory where action plugins live in the `action` folder. And yes, that mix of singular and plural names is making me go crazy, but there is not much to do here since this is what Ansible prescribes ;)

tests/unit/action/test_bonsai_asset.py
tadeboro: What I meant to say was `# Optional values are not checked for type-correctness if they are missing.`. And now that I am thinking about this, I also forgot to include a test to make sure optional value fails the validation if the type is not correct.

tests/unit/action/test_bonsai_asset.py
tadeboro: I used this form because this is how the actual code is supposed to handle errors. Ansible stuff really does not care what erorr exactly was raised, all it needs to do is make sure it does not bubble up from the plugin and instead gracefully construct a return value.

plugins/module_utils/errors.py
```diff
@@ -21,3 +21,7 @@ class SyncError(Error):
 
 class SensuError(Error):
     """ Error that signals problems with Sensu Go web API. """
+
+
+class BonsaiError(Error):
+    """ Error that signals problems with Bonsai assets. """
```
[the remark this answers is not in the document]
tadeboro: The general idea behind the exception naming scheme was to make it easy to identify what part is responsible for the failure. For example, `BonsaiError` indicates that something with external Bonsai service went haywire and `SensuError` should point to interactions with Sensu Go backend. What I do find out of place is that `SyncError`, since this clearly belongs to the `SensuError` category.

Maybe we should structure them into only three groups:

 1. `HttpError` indicates connection problems (cannot connect and other low-level stuff).
 2. `SensuError` indicates logical problems with Sensu API.
 3. `BonsaiError` indicates logical problems with Bonsai API.

This, combined with a custom `__str__` method on the `Error` class, should give us a much better idea where the error stems from.

That being said, maybe we can fix this in a follow-up PR? This change will not affect the public API in any way, so this should be safe to postpone a bit.

tests/unit/action/test_bonsai_asset.py
mancabizjak: Oh, right, I see. I'm good then :+1:

plugins/module_utils/errors.py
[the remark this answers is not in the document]
mancabizjak: Hmm, indeed, if I see `SyncError`, then I would also expect `AuthError` and the like. And if this were the case, we would quickly find ourselves in a situation where we have different error type for alsmost every error. Grouping errors with respect to domains (Bonsai, Sensu) and removing error types that are off rather than adding new ones, does make sense.

Yes, let's keep it as it is for now - regardless of the error types that we're currently using, the error messages themselves are descriptive enough. We can come back to it at a later time and revisit the naming scheme.

plugins/module_utils/bonsai.py
```diff
@@ -0,0 +1,69 @@
...
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    errors, http,
+)
+
+
+def get(path):
+    url = "https://bonsai.sensu.io/api/v1/assets/{0}".format(path.lstrip("/"))
```
mancabizjak: Do we need the `lstrip` part? Apart from the unit tests where `path` comprises a leading `/`, `get` is always called with path argument without it. I mean, it certainly doesn't hurt if we keep it, but maybe it's not required.

tadeboro: We do not. I do not even remember adding it, so let us simplify this.

mancabizjak:
I'm quite happy with the updates. And I also think this module brings a lot of added value for the end-user, which is great :+1:. Thanks @tadeboro!

PR #110, opened 2019-11-26
Refactor URL path construction

tadeboro:
Up until now, we manually constructed paths for Sensu API client. And while this worked well for us, it is still a bit unsafe and can lead to some hard-to-debug situations.

In this commit, we introduced the `build_url_path` function that performs any escaping while constructing the final path. And while we were at it, we also made sure this function can handle missing parts of the path. With this additional functionality, we were able to drop quite a few if clauses in the code.

We also introduced quoting into the Sensu API client itself since the namespace is another piece of information that we retrieve from the Ansible users and should be treated with some care.

Note that we did not change the integration tests. This gives us some confidence that we are on the right track with this change.


tests/unit/modules/test_silence.py
```diff
@@ -26,7 +26,7 @@ def test_minimal_silence_parameters_check(self, mocker):
 
         state, _client, path, payload, check_mode = sync_mock.call_args[0]
         assert state == 'present'
-        assert path == '/silenced/*:check'
+        assert path == '/silenced/%2A%3Acheck'  # %2A = *, %3A = :
```
mancabizjak: Happy to see comments like these :)

tests/unit/modules/test_silence_info.py
```diff
@@ -35,7 +35,8 @@ def test_get_single_silence(self, mocker):
             silence_info.main()
 
         _client, path = get_mock.call_args[0]
-        assert path == "/silenced/subscription:*"
+        print(path)
```
aljazkosir: Is this a leftover from debugging ;)?

tadeboro: I never debug, my code works on the first try ;) Nice catch indeed. Thanks @aljazkosir.

aljazkosir:
Thanks @tadeboro

tadeboro:
I rebased this onto the master branch to get the CI green.

PR #111, opened 2019-11-26
Bonsai asset action plugin

tadeboro:
This is an alternative implementation of functionality in #108. Note that this is not completely done yet since we are missing tests for action plugin, but I wanted to show the implementation before I start polishing it and wasting time if we decide to ignore it.


tadeboro:
I am closing this PR and moving the changes to #108 where we have API debate.

PR #112, opened 2019-11-27
Use custom, prebuilt images for integration testing

tadeboro:
We used stock docker images for our integration tests, which was not optimal since we modified each image before running tests on it. But if we put those inefficiencies aside, things worked great.

But today the builds started failing because the official Ubuntu Trusty image does not work without subscription anymore. This prompted us to revisit the image situation once again and fix it for good.

What we ended up doing is building our own images that contain everything tests need to run successfully. They are hosted under the https://hub.docker.com/u/xlabsi organization on the docker hub.

With those images ready for use, we were able to flip the pre_build_image switch to true in the Molecule's docker driver.

And while we were modifying the image definitions in molecule definitions, we also added remaining missing versions of Sensu to our test suite. We now test almost all supported versions of Sensu, with the sole exception of version 5.15.0 that has a bug in its API.


tests/integration/base.yml
```diff
@@ -62,15 +62,36 @@ verifier:
...
+    command: *cmd
   - name: v5.13.1
-    image: sensu/sensu:5.13.1
+    image: xlabsi/sensu:5.13.1
+    pre_build_image: true
+    pull: true
     command: *cmd
   - name: v5.12.0
-    image: sensu/sensu:5.12.0
+    image: xlabsi/sensu:5.13.0
```
mancabizjak:
```suggestion
    image: xlabsi/sensu:5.12.0
```

tadeboro: I forgot to change the name. Sensu 5.12 is not supported anymore so there is no need to test it.

mancabizjak:
Thank you @tadeboro, for all the effort that went into saving our CI so promptly.

PR #113, opened 2019-11-28
Clean up obsolete modules, tests and docs

mancabizjak:
Commits in this PR mostly delete files - old unit and integration tests, and the `sensu_go_*` modules that were not updated to the new structure so far.

Documentation fragments are also cleaned up a little. Two fragments are removed, including `doc_fragments/base.py`, which is essentially the same as `doc_fragments/auth.py`. A couple of modules still kept a reference to `base.py`, which we corrected here.


mancabizjak:
Oops, forgot to rebase.

mancabizjak:
Thanks, I removed `base.py` as well. 

> * `plugins/doc_fragments/info.py` - this was used by the legacy info modules

Actually it's still in use by all info modules. `info.py` doc fragment is almost the same as `name.py` one, except that `name` is required there, but it isn't required for the info modules.

tadeboro:
> > * `plugins/doc_fragments/info.py` - this was used by the legacy info modules
> 
> Actually it's still in use by all info modules. `info.py` doc fragment is almost the same as `name.py` one, except that `name` is required there, but it isn't required for the info modules.

You are right. I forgot that we have two different specifications for `name` parameter.

PR #120, opened 2019-12-03
Module documentation fixes

tadeboro:
This is another installation in the "Let us get 1.0 out the door" saga ;) As always, detailed descriptions of changes are in the commit messages.

### DIFF
--- a/plugins/modules/namespace_info.py
+++ b/plugins/modules/namespace_info.py
@@ -29,7 +29,11 @@ version_added: "1.0"
 extends_documentation_fragment:
   - sensu.sensu_go.auth
 notes:
-  - Parameters C(auth.namespace) is ignored in this module.
+  - Parameters I(auth.namespace) is ignored in this module.
+  - Currently, it is not possible to retrieve information about a single
+    namespace because namespace is not much more than a name itself.
+seealso:
+  - module: namespace
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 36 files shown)


plugins/modules/silence.py
```diff
@@ -32,6 +32,8 @@
   - sensu.sensu_go.state
```
mancabizjak: Can you please add a note documenting that at least one of (`subscription`, `check`) is required?

plugins/modules/silence_info.py
```diff
@@ -29,6 +29,8 @@
 version_added: "1.0"
 extends_documentation_fragment:
   - sensu.sensu_go.auth
+seealso:
+  - module: silence
 options:
   subscription:
     description:
```
mancabizjak: Would you please update also these lines below:
* L48: List *all* Sensu silence entries
* L52: Fetch *a* specific silence....

plugins/modules/socket_handler.py
```diff
@@ -78,6 +82,11 @@
     type: udp
     host: 10.0.1.99
     port: 4444
+
+- name: Delete handler
```
mancabizjak:
```suggestion
- name: Delete a handler
```

plugins/modules/namespace_info.py
```diff
@@ -63,7 +67,7 @@ def main():
     path = '/namespaces'
 
     try:
-        namespaces = utils.get(client, path)
+        namespaces = utils.prepare_result_list(utils.get(client, path))
```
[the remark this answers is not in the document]
tadeboro: Actually, this is an old remnant that I did not intend to commit. We cannot fetch a single namespace, which means that we do not need to filter the result of the `GET` call.

I know that having this here does not hurt us, but I would rather keep it out because this allows our integration tests to catch incompatible changes in the Sensu's web API.

PR #121, opened 2019-12-05
Move namespace parameter out of auth data

tadeboro:
When we were designing a general API layout, we got at least one thing wrong: we bundled namespace data with backend authentication parameters. And what is even more unfortunate is that the layout worked just fine because most of the time, you operate on a single
Sensu Go namespace. This is why we detected this design flaw only when we started writing the end-user documentation.

The fix is simple in theory: just move the namespace parameter. In practice, this change was annoying as hell because we were changing the public API of our modules, which translates into a whole bunch of broken tests.

The main steps we did when moving the namespace parameter were:

 1. Remove all notions of the namespace from the Sensu HTTP client.
 2. Move URL construction into a separate helper method.
 3. Update parameter descriptions for each module.
 4. Fix URL construction in each module.
 5. Update tests to make sure namespace parameter is properly tested.

And while we were changing the Sensu HTTP client, we also removed the hardcoded /api/core/v2/ prefix. The main reason for doing this is future extensibility. With this change in place, we can interact with Sensu endpoints that live outside the previously fixed prefix.

And if you want to start yelling YAGNI right now, let us calm you down. It is not question of if we will need those changes, it is a question of when we will need them.


plugins/modules/cluster_role_info.py
```diff
@@ -70,9 +70,10 @@ def main():
         ),
```
mancabizjak: Please also remove _Parameter C(auth.namespace) is ignored in this module_ from the docs for this module.

tests/unit/modules/test_check_info.py
```diff
@@ -17,13 +17,14 @@ class TestSensuGoCheckInfo(ModuleTestCase):
...
         get_mock.return_value = [1, 2, 3]
-        set_module_args()
+        set_module_args(namespace="my")
 
         with pytest.raises(AnsibleExitJson) as context:
             check_info.main()
 
         _client, path = get_mock.call_args[0]
-        assert path == "/checks"
+        print(path)
```
mancabizjak: Oopsie :)

tests/unit/modules/test_user_info.py
```diff
@@ -23,7 +23,7 @@ def test_get_all_users(self, mocker):
             user_info.main()
 
         _client, path = get_mock.call_args[0]
-        assert path == "/users"
+        assert path == "/api/core/v2/users"
         assert context.value.args[0]["objects"] == [1, 2, 3]
 
     def test_get_single_mutator(self, mocker):
```
mancabizjak: Ugh, I just noticed this.
I know this is totally unrelated to the changes you made in this PR, but could you please just rename it to `test_get_single_user`?

PR #125, opened 2019-12-16
Add external datastore modules

### LINKED ISSUE #119: Feature Request: Support for 3rd party datastore
Presently, the collection doesn't allow for a 3rd party datastore to be configured. The guide for this is located here: https://docs.sensu.io/sensu-go/5.15/guides/scale-event-storage/.

tadeboro:
Datastore module and its info counterpart know how to deal with external datastores.

The payload that the Sensu storage API accepts is a bit different from the payload that core/v2 API endpoints expect, but we kept the user-facing interface similar to all other modules. This way, playbook authors can use their knowledge about existing modules when dealing with data storage.

Fixes #119

### DIFF
--- a/plugins/modules/datastore.py
+++ b/plugins/modules/datastore.py
@@ -1,0 +1,161 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2019, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["stableinterface"],
+    "supported_by": "certified",
+}
+
+DOCUMENTATION = '''
+module: datastore
+author:
+  - Manca Bizjak (@mancabizjak)
+  - Tadej Borovsak (@tadeboro)
+short_description: Manage Sensu external datastore providers
+description:
+  - Add or remove external datastore provider.
+  - For more information, refer to the Sensu documentation at
+    U(https://docs.sensu.io/sensu-go/latest/reference/datastore/).
+version_added: "1.1"
+extends_documentation_fragment:
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.name
+  - sensu.sensu_go.state
+seealso:
+  - module: datastore_info
+options:
+  dsn:
+    description:
+      - Attribute that specifies the data source names as a URL or
+        PostgreSQL connection string. See the PostgreSQL docs for more
+        information about connection strings.
+    type: str
+  pool_size:
+    description:
+      - The maximum number of connections to hold in the PostgreSQL connection
+        pool.
+    type: int
+notes:
+  - Currently, only one external datastore can be active at a time. The module
+    will fail to perform its operation if this would break that invariant.
+'''
+
+EXAMPLES = '''
+- name: Add external datastore
+  datastore:
+    name: my-postgres
+    dsn: postgresql://user:secret@host:port/dbname
+
+- name: Remove external datastore
+  datastore:
+    name: my-postgres
+    state: absent
+'''
+
+RETURN = '''
+object:
+    description: object representing external datastore provider
+    returned: success
+    type: dict
+'''
+
+from ansible.module_utils.basic import AnsibleModule
+
+from ansible_collections.sensu.sensu_go.plugins.module_utils import (
+    arguments, errors, utils,
+)
+
+API_GROUP = "enterprise"
+API_VERSION = "store/v1"
+
+
+def _do_differ(remote, payload):
+    return utils.do_differ(remote["spec"], payload["spec"])
+
+
+def sync(state, client, list_path, resource_path, payload, check_mode):
+    datastore = utils.get(client, resource_path)
+
+    # When we are deleting stores, we do not care if there is more than one
+    # datastore present. We just make sure the currently manipulated store is
+    # gone. This makes our module useful in "let us clean up the mess"
+    # scenarios.
+    if state == "absent" and datastore is None:
+        return False, None
+
+    if state == "absent":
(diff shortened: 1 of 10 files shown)


plugins/modules/datastore.py
```diff
@@ -0,0 +1,161 @@
...
+
+    if check_mode:
+        return True, payload
+    utils.put(client, resource_path, payload)
+    return True, utils.get(client, resource_path)
+
+
+def main():
+    required_if = [
+        ("state", "enabled", ["dsn"])
```
mancabizjak: I believe this should be:
```suggestion
        ("state", "present", ["dsn"])
```

tests/unit/modules/test_datastore.py
```diff
@@ -0,0 +1,253 @@
...
+        client.put.assert_not_called()
+
+
+class TestDatastore(ModuleTestCase):
+    def test_minimal_datastore_parameters(self, mocker):
+        sync_mock = mocker.patch.object(datastore, "sync")
+        sync_mock.return_value = True, {}
+        set_module_args(
+            name="test_datastore",
+            dsn="my-dsn",
```
mancabizjak: According to the current argument spec, the minimal set of parameters includes only `name`. If you omitted `dsn` here, the validation wouldn't prevent you from doing that because currently `dsn` is not required for state = present but for state = enabled.

tests/integration/molecule/module_datastore/playbook.yml
```diff
@@ -0,0 +1,122 @@
...
+    - name: Retrieve empty list of external datastores
+      datastore_info:
+        auth: &auth
+          url: http://localhost:8080
+      register: result
+
+    - assert:
+        that:
+          - result.objects == []
+
```
mancabizjak: Can you add another test here, that ensures validation kicks in, to ensure we cannot create a datastore without a missing `dsn`?

plugins/modules/datastore.py
```diff
@@ -0,0 +1,161 @@
...
+}
+
+DOCUMENTATION = '''
+module: datastore
+author:
+  - Manca Bizjak (@mancabizjak)
+  - Tadej Borovsak (@tadeboro)
+short_description: Manage Sensu external datastore
+description:
+  - Activate or deactivate external datastore.
```
mancabizjak:
```suggestion
  - Add or remove external datastore provider.
```

plugins/modules/datastore.py
```diff
@@ -0,0 +1,161 @@
...
+    "status": ["stableinterface"],
+    "supported_by": "certified",
+}
+
+DOCUMENTATION = '''
+module: datastore
+author:
+  - Manca Bizjak (@mancabizjak)
+  - Tadej Borovsak (@tadeboro)
+short_description: Manage Sensu external datastore
```
mancabizjak:
```suggestion
short_description: Manage Sensu external datastore providers
```

plugins/modules/datastore_info.py
```diff
@@ -0,0 +1,89 @@
...
+    "status": ["stableinterface"],
+    "supported_by": "certified",
+}
+
+DOCUMENTATION = """
+module: datastore_info
+author:
+  - Manca Bizjak (@mancabizjak)
+  - Tadej Borovsak (@tadeboro)
+short_description: List external Sensu datastores
```
mancabizjak:
```suggestion
short_description: List external Sensu datastore providers
```

plugins/modules/datastore_info.py
```diff
@@ -0,0 +1,89 @@
...
+  register: result
+
+- name: Do something with result
+  debug:
+    msg: "{{ result.objects.0.dsn }}"
+"""
+
+RETURN = """
+objects:
+  description: list of external Sensu datastores
```
mancabizjak:
```suggestion
  description: list of external Sensu datastore providers
```

plugins/modules/datastore.py
```diff
@@ -0,0 +1,161 @@
...
+
+- name: Remove external datastore
+  datastore:
+    name: my-postgres
+    state: absent
+'''
+
+RETURN = '''
+object:
+    description: object representing external datastore
```
mancabizjak:
```suggestion
    description: object representing external datastore provider
```

tests/integration/molecule/module_datastore/playbook.yml
```diff
@@ -0,0 +1,122 @@
...
+    - name: Enable external datastore with default pool_size
```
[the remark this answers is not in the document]
tadeboro: Initially, I added a default value for the `pool_size` parameter, then I wrote the tests, then I remembered that we do not use default values for optional parameters and let the backend to deal with them, then I forgot to update the test name ;)

tests/unit/modules/test_datastore.py
tadeboro: I completely forgot that we have two minimal sets of parameters. I added another test for `state == "absent"`. Great catch here.

plugins/modules/datastore.py
tadeboro: These are remnants of my previous incarnations of this module where I blindly assumed that external datastore is either enabled or disabled. In theory, things are a bit more flexible, but for all practical purposes, my initial assumption was not completely wrong ;)

tadeboro:
Let me quickly explain the "why" behind the implemented sync logic. Basically, Sensu Go only supports [a single external datastore](https://github.com/sensu/sensu-go-ansible/issues/119#issuecomment-566316618), but as you noticed, the backend does not enforce this. Because this was a premium feature up until Sensu Go 5.15 and is still not part of the open-source code, I cannot easily verify if adding a second datastore to the backend breaks anything. So I played it safe here and placed the restrictions in the code.

It would be much simpler just to document this, but I felt that making sure we do not brick the user's Sensu installation justifies this additional code.

mancabizjak:
Ah, I see :slightly_smiling_face:  Thanks for the link, clarification, and of course for the additional modules! Looks great.

mancabizjak:
There seems to be a typo in the integration test on L20. I believe we need to have `auth: *auth` instead of `auth: &auth`. Otherwise we're good to go :+1:

PR #128, opened 2019-12-19
Make sure backend role works on Sensu Go >= 5.16

tadeboro:
Details are described in the commit messages.

Depends on #127


mancabizjak:
Thanks, @tadeboro, looks great. You can rebase the branch and remove the "draft" status.

One more thing. Can you update the time for *role_backend_default* in _scenario.times_? On my system, the speedup is indeed quite notable.

tadeboro:
I think I fixed all my screwups, so this pull request should be ready to go.

mancabizjak:
Would you please also update L18 in `docs/source/roles/install.rst` (see my comment above)? It's just to stay consistent with the changes you made to the examples for the install role. Then we're good to go :+1:

PR #130, opened 2019-12-22
add namespace support to bonsai_asset

jakeo:
nice work on the collection!

I came across this error when using `bonsai_asset` with a non-"default" namespace. This is hopefully the correct fix.

### DIFF
--- a/plugins/action/bonsai_asset.py
+++ b/plugins/action/bonsai_asset.py
@@ -31,7 +31,7 @@ def validate(name, args, required, typ):
 class ActionModule(ActionBase):
 
     _VALID_ARGS = frozenset(
-        ("auth", "name", "version", "rename", "labels", "annotations")
+        ("auth", "name", "version", "namespace", "rename", "labels", "annotations")
     )
 
     def run(self, _tmp=None, task_vars=None):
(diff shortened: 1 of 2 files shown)


plugins/action/bonsai_asset.py
```diff
@@ -67,6 +67,7 @@ def validate_arguments(args):
 
         validate("name", args, required=True, typ=str)
         validate("version", args, required=True, typ=str)
+        validate("namespace", args, required=False, typ=str)
```
tadeboro: This line should be removed since we are not using the `namespace` parameter in our code. As the comment in this function states, the asset module will validate the parameters of this kind (`auth` was the only one until now).

jakeo:
> Great catch, @jakeo! And a good first stab at fixing this bug I introduced when I was moving the `namespace` parameter in #121.
> 
> Would you be so kind and add a test for your code? You can start by copying the test for the [auth parameter](https://github.com/sensu/sensu-go-ansible/blob/master/tests/unit/action/test_bonsai_asset.py#L199) and adjusting it for the `namespace` parameter. Thanks in advance.

👍 - ran the new unit w/ `pytest` - hope this helps! happy holidays 🎁

PR #136, opened 2020-01-07
Tighten static analysis

tadeboro:
Import process on Ansible Galaxy runs some static checks on the code that `ansible-test sanity` does not. Changes in this PR add those check to our `sanity` make target.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,22 @@
+# This configuration mirrors the configuration that is used when linting
+# collections on import into Ansible Galaxy.
+
+# TODO(@tadeboro): Some of our lines are almost twice the optimal reading
+# length of 60-80 chars. It would be great to get line lengths down to 79, but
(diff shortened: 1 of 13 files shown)


.flake8
```diff
@@ -0,0 +1,22 @@
+# This configuration mirrors the configuration that is used when linting
+# collections on import into Ansible Galaxy.
+
+# TODO(@tadeboro): Some of our lines are almost twice the optimal reading
+# length of 60-80 chars. It would be great to get line lengths down to 79, but
+# this is not feasible at the moment since we have cca. 300 lines that are
+# longer.  What we need is a way of preventing new long lines of getting into
```
mdolin: Extra space after "longer.". For text sanity we should probably remove that extra space.

tadeboro:
I addressed all the comments. Once the tests are green, we should be ready to go.

PR #137, opened 2020-01-07
Strip empty lines from Sensu configuration files

tadeboro:
While it is true that the generated configuration files should not be modified manually, they might still get inspected from time to time. And in those cases, having a nice, readable configuration file can
make a significant difference in usability.

But because we do not want to bundle all our template conditional logic into one continuous block that looks like a core dump, we implemented the cleanup by using Jinja's trim modifiers to code blocks.

This way we can almost have a cake and eat it too: the template is readable from the developer's perspective and the generated configuration is concise and clean.


matejart:
Thank you, @tadeboro, this looks good. 👍

PR #146, opened 2020-01-17
Prepare collection for 1.2.0 release

tadeboro:
This PR also includes some changes that should be part of the 1.1.1 release, but I forgot to include them. I did split them into a separate commit in order to keep things as clean as possible.


mancabizjak:
LGTM, thanks @ŧadeboro.

PR #147, opened 2020-01-20
Scenario docs

tadeboro:


jspaleta:
Okay the installing backend and monitoring hosts scenario documentation and examples look good.  
At most minor changes needed there.

The "Managing resources" scenario documentation needs some work. Maybe we remove that one for now?  The playbook works though.

If we keep it I'd suggest we reorder the scenarios and add the managing resources after the agent provisioning scenario.

tadeboro:
Closing since last interaction was 2 years ago.

Issue #148, opened 2020-01-22
Missing Debian.yml in for install role


jspaleta:
**Bug description**
Missing roles/install/vars/Debian.yml

Trying to run agent install against a debian host and the playbook errors out.

**How to reproduce the bug**
1. fire up debian stretch vagrant host
2. ansible-playbook -i inventory.yaml playbook.yaml 

**Expected result**
playbook should run

**Actual result**
```
TASK [install : Include distro-specific vars (Debian)] ******************************************************************************************************************
fatal: [192.168.50.11]: FAILED! => {"ansible_facts": {}, "ansible_included_var_files": [], "changed": false, "message": "Could not find or access 'Debian.yml'\nSearched in:\n\t/home/jspaleta/.ansible/collections/ansible_collections/sensu/sensu_go/roles/install/vars/Debian.yml\n\t/home/jspaleta/.ansible/collections/ansible_collections/sensu/sensu_go/roles/install/Debian.yml\n\t/home/jspaleta/.ansible/collections/ansible_collections/sensu/sensu_go/roles/agent/vars/Debian.yml\n\t/home/jspaleta/.ansible/collections/ansible_collections/sensu/sensu_go/roles/agent/Debian.yml\n\t/home/jspaleta/.ansible/collections/ansible_collections/sensu/sensu_go/roles/install/tasks/apt/vars/Debian.yml\n\t/home/jspaleta/.ansible/collections/ansible_collections/sensu/sensu_go/roles/install/tasks/apt/Debian.yml\n\t/home/jspaleta/sources/orgs/sensu/sensu-go-ansible/docs/examples/scenarios/monitoring_hosts/vars/Debian.yml\n\t/home/jspaleta/sources/orgs/sensu/sensu-go-ansible/docs/examples/scenarios/monitoring_hosts/Debian.yml on the Ansible Controller.\nIf you are using a module and expect the file to exist on the remote, see the remote_src option"}
```

**Component versions**
 - Ansible Version: 2.9.3
 - Sensu Go Collection Version 1.2.0
 - Sensu Go Backend Version  5.16.1

**Additional context**
Using playbooks and Vagrantfiles from scenario-docs branch
Adding Debian.yml at location indicated in the error solves the problem.

tadeboro:
The fix for this bug is already in #147 (I noticed the lack of Debian support while working on the docs), we just need to add a platform to our tests and extract the changes into a separate PR in order to keep things well tested and tidy.

[...]

PR #151, opened 2020-01-22
Add Debian support

tadeboro:
Changes in this PR add support for Debian. As in order to keep our codebase as clean as possible, we took the chance and remove the final remaining python test code from the integration tests.


tadeboro:
Fixes #148

mancabizjak:
Thank you @tadeboro. I'm also very glad to see the role integration tests fully consistent.

PR #153, opened 2020-01-23
working 3 member sensu cluster install playbook  and cleanup playbook

jspaleta:
Sensu backend cluster scenario.

I'll be continuing to add to this.

Here's what is working right now.

playbook.yml  will spin up a working cluster
cleanup_playbook.yml will disable the cluster and empty out each member's etcd store making it possible to quickly re-provision a new cluster from scratch (skipping the package download step)


jspaleta:
Added a loadblancer vm into the scenario.  Provisioned via simple inline script in the Vagrant file.

I didn't want to overly complicate the Sensu focused playbook with the loadbalancer provisioning.

jspaleta:
I'm going to move this off into a new sub-directory under examples instead of trying to have it be a documented scenario.  We can perhaps point to it as and advanced reference example

jspaleta:
Closing this PR out, as its replaced by #213

PR #154, opened 2020-01-24
Add API key authentication support

tadeboro:
Sensu Go supports API key authentication since version 5.15.0 and it is about time we also add it as a valid method of authentication.

In order not to break the backward compatibility, we made the API key optional. If the API key is not present in the parameters, we proceed with the username and password login. If the API key is set, we validate it and in the case of failure report the error back to the user.

Note that if the API key is present nut invalid, we do not try to authenticate using username and password. We wrote the code this way on purpose, since silently ignoring authentication failure with something that the user believes is valid is not an option.

### DIFF
--- a/docs/source/modules.rst
+++ b/docs/source/modules.rst
@@ -7,31 +7,114 @@ the same pattern as much as possible.
 The API of each module is composed of two parts. The *auth* parameter contains
 the pieces of information that are related to the Sensu Go backend that the
 module is connecting to. All other parameters hold the information related to
-the resource that we are operating on. A typical task would look like this:
+the resource that we are operating on.
+
+
+Authentication parameters
+-------------------------
+
+Each module has an *auth* parameter that holds the following information about
+the Sensu Go backend:
+
+1. The **url** key holds the address of the Sensu Go backend. If this key is
+   not present in the task's definition, Ansible will consult the *SENSU_URL*
+   environment variable and, if the variable is not set, use the default value
+   of ``http://localhost:8080``.
+2. The **user** and **password** keys contain the credentials that the module
+   will use when connecting to the backend. It not present, Ansible will try
+   to look up the *SENSU_USER* and *SENSU_PASSWORD* environment variables,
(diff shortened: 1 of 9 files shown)


tests/unit/module_utils/test_client.py
```diff
@@ -13,60 +13,98 @@
...
         request = mocker.patch.object(http, "request")
         request.return_value = http.Response(200, '{"access_bla": "token"}')
 
         with pytest.raises(errors.SensuError, match="token"):
-            client.Client("http://example.com/", "user", "pass").token
+            client.Client(
+                "http://example.com/", "user", "pass", None,
+            ).auth_header
+
+    def test_login_failure_api_Key_bad_status(self, mocker):
```
mancabizjak:
```suggestion
    def test_login_failure_api_key_bad_status(self, mocker):
```

tadeboro:
I think I resolved most of the issues with this PR (removed typos, added tests on non-api-key-enabled backend and expanded the authentication part into a separate section).

docs/source/modules.rst
```diff
@@ -7,31 +7,114 @@ the same pattern as much as possible.
...
+authenticate using the API key. If the API key is not set, Ansible will
+fallback to using the *user* and *password* values for authentication. What
+this basically means is that there are two valid sets of *auth* parameters:
 
 .. code-block:: yaml
 
-   - name: Make sure asset is present
-     asset:
+   - name: Use API key authentication
+     sensu.sensu_go.asset:
```
mancabizjak: Prefixing with namespace & collection name is totally fine, however we now have some examples in this file that use this notation and others that assume `collections: [ sensu.sensu_go ]` is specified in the play. However feel free to ignore this nitpick, since it's true that we demonstrate both ways of referring to collection content in the quickstart document.

tadeboro: I will unify the notation. Let us not give our users yet another reason to be confused ;)

tadeboro:
I think I resolved all of the issues, take two ;)

mancabizjak:
Looks great, thanks for the effort @tadeboro!

PR #159, opened 2020-02-03
Webinar content

tadeboro:
This is an almost-final technical demo for the Sensu + Ansible webinar. What is missing is a custom asset that we would upload to the asset server and a check that would utilize it.

/cc @jspaleta


tadeboro:
The basic scenario is a copy of the getting started content from the official Sensu Go site. This demo serves as a demonstration on how to automate the manual steps that we went through before the technical demo. We will talk about the inventory here, how tasks map to hosts, and where are modules that interact with the Sensu Go API executed.

tadeboro:
The cluster scenario contains:

 1. a 3-member Sensu Go backend cluster,
 2. a load balancer before the backend cluster,
 3. an asset server that contains custom assets,
 4. an agent VM that will use a custom asset.

The general idea behind this demo is to demonstrate some of the advanced capabilities of Sensu Go and how that maps to the Ansible playbooks.

jspaleta:
I think we can close this out.
The stripped down cluster scenario PR is probably adequate addition to the documented scenarios

tadeboro:
Closing since it has been 2 years since last comment.

PR #173, opened 2020-03-30
Add support for CentOS/RHEL 8

tadeboro:
The latest version of the OS from Red Hat uses dnf to manage system packages. Fortunately, Ansible already contains appropriate modules so adding support was relatively painless.

Do note that we copied the contents of the prepare.yml file from the yum repo verbatim. We did this on purpose because we do not want dnf and yum functionality to be tied together unnecessarily.


tadeboro:
I know we could use a test or two more, and I intend to add them as soon as we resolve the Molecule issue at https://github.com/ansible-community/molecule/pull/2631

mancabizjak:
Thank you @tadeboro.

PR #176, opened 2020-04-18
Remove ubuntu 19 4

tadeboro:
Ubuntu 19.04 reached EOL and the apt repositories were shut down.

And while we were updating the tests, we also bumped the parallelism level to 8, reducing the test time almost in half.


mancabizjak:
> And while we were updating the tests, we also bumped the parallelism level to 15, reducing the test time almost in half.

Just checking, you probably meant 8 (judging from the code changes), right? :) But based on a quick CircleCI documentation check, it seems like we could potentially go up to 16?

In any case, lovely to see test run way faster! :+1:

tadeboro:
On our current plan, we can bump parallelism to 200 if we wanted, but I learned the hard way that the remote docker capabilities do not scale in proportion to the number of parallel workers available. I initially bumped the parallelism level to 16, but the operations started to timeout and I was forced to find an acceptable middle ground.

In order to keep the test times sane and prevent them from sporadically failing, I did take some even more drastic measures in the last commit of https://github.com/sensu/sensu-go-ansible/pull/180 where I limited the testing to one version per minor release. Otherwise, we would need to run tests against 9 or 10 versions instead of 4 for very limited gain.

mancabizjak:
Thanks for the additional insight @tadeboro, and for your tireless work on keeping the test suite green, up-to-date, and as fast as possible.

PR #178, opened 2020-04-19
Refactor backend role

### LINKED ISSUE #170: Initialize backend fails in check mode
**Bug description**
When running through the backend playbook in ansible checkmode Initialize backend fails because `init_command_test` is undefined.

**How to reproduce the bug**
1. Run the command `ansible-playbook -i inv ... -C`.

**Expected result**
This step should just skip like the `Check for sensu-backend init command` does

**Actual result**
Ansible exits prematurely
```
FAILED! => {"msg": "The conditional check 'init_command_test.rc == 0' failed. The error was: error while evaluating conditional (init_command_test.rc == 0): 'dict object' has no attribute 'rc'\n\nThe error appears to be in '.../collections/ansible_collections/sensu/sensu_go/roles/backend/tasks/main.yml': line 80, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Initialize backend\n  ^ here\n"}
```

**Component versions**
- Python Version: 3.6.9
 - Ansible Version: 2.9.6
 - Sensu Go Collection Version: 1.3.0

tadeboro:
The primary reason for this PR is our failure to be able to execute the backend role in check mode. In short: because the command module does not support check mode, Ansible skips it when running in check mode, and that caused one of the tasks relying on the registered variable to fail.

We were able to fix this with relative ease because the command that we are executing on the remote host is safe to run even in check mode (it does not modify the system at all).

We refactored the role mainly because this way, the backend role is more flexible. And the fact that we can now test the backend role more efficiently and in greater detail is also not to be sniffed at.

~Depends on #177~
Fixes #170


tadeboro:
I prefer to keep the checks in a single playbook if possible. This makes it easy to follow the "flow" of the test scenario: we do something and then immediately check the results. This is how our modules are tested and this is how all our roles were tested until this PR. This way, we can keep our test sequence simple: create -> converge -> destroy.

I split the default backend test scenario into multiple files because I introduced another step into the test sequence: the check step. In the check phase, the molecule will run the `converge.yml` playbook in check mode, and it felt a bit wrong to test if our stat and assert tasks that have nothing to do with the role work in check mode.

PR #186, opened 2020-05-17
Fake event module result

tadeboro:
Sensu Go backend processes events in an asynchronous fashion, which makes it almost impossible to retrieve the processed event shortly after it has been sent to the backend.

Currently, after we submit an event, we query the backend for the last event. What we get back is one of the following three options:

 1. nothing, if our event is the first event ever and the backend did not manage to process it yet;
 2. previous event, if there is at least one preexisting event on the backend and our event has not been processed yet;
 3. our actual event, if the backend processed our event fast enough.

In the real world setting, the third scenario almost never happens, which means that our event module is returning bogus results most of the time.

To rectify the situation, we now simply return back the payload we sent to the backend. This ensures that we return back relevant pieces of information instead of some random or event or even none.

### DIFF
--- a/plugins/modules/event.py
+++ b/plugins/modules/event.py
@@ -154,7 +154,7 @@ EXAMPLES = '''
 
 RETURN = '''
 object:
-  description: object representing Sensu event
+  description: object representing Sensu event (deprecated)
   returned: success
   type: dict
 '''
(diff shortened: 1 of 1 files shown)


tadeboro:
An alternative solution is presented in #185

tadeboro:
We decided to go through the deprecation cycle with the result modification. This PR introduces the deprecation and a temporary workaround, and we repurposed the #185 to clean the deprecation and remove the object result field.

PR #199, opened 2020-06-26
Make sure the check module is idempotent

### LINKED ISSUE #198: Tasks are not impodent
At least the following tasks are not impodent:

- bonsai_asset
- filter
- pipe_handlers
- handler_set
- check
- entity

This results in tasks reporting the `changed` state when not changed, while they should report `ok` state.

tadeboro:
The payload for creating a check is large and contains quite a few lists that serve as a set of things. All this makes comparing the payloads a bit tougher than for the rest of the resources.

Another source of problems is the backend's tendency to treat the null and empty lists as equivalent, adding yet another place where equality checks can go wrong.

Changes in this commit try to fix all of the idempotency issues we could think of, but there is a slight possibility that we still missed something. But things should be in much better shape now.

Fixes #198

### DIFF
--- a/tests/integration/molecule/module_check/converge.yml
+++ b/tests/integration/molecule/module_check/converge.yml
@@ -197,3 +197,58 @@
     - assert:
         that:
           - result.objects == []
+
+    - name: Create check for idempotency test of complex fields
+      check:
+        name: complex
+        command: sleep 10
+        interval: 10
+        subscriptions:
+          - sub1
+          - sub2
+        handlers: []
+        proxy_requests:
+          entity_attributes:
+            - "entity.entity_class == 'proxy'"
+            - "entity.entity_class == 'demo'"
+        check_hooks:
+          warning:
+            - h1
+            - h2
+            - h3
+          error:
+            - h4
+            - h2
+        env_vars:
+          var1: val1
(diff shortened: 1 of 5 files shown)


tests/integration/molecule/module_check/converge.yml
```diff
@@ -197,3 +197,58 @@
...
+    - name: Create check for idempotency test of complex fields
+      check:
+        name: complex
+        command: sleep 10
+        interval: 10
+        subscriptions:
+          - sub1
+          - sub2
+        handlers: []
+        runtime_assets: []
```
mancabizjak, line 210: Maybe we could omit one of these parameters in the first task and then set its value to `[]` in the idempotency check task, to test the reverse situation. This should work, right?

```
- name: Create check for idempotency test of complex field
  check:
    # handlers are null
    runtime_assets: []
```

and then

```
- name: Test for idempotency test of complex fields
   check:
     handlers: []
     # runtime_assets are null
```

tadeboro: I thought I did this already. I guess not, Great catch.

tadeboro:
I updated the integration test as suggested.

PR #200, opened 2020-06-26
Sensu Go 5.21.0 updates

tadeboro:
Changes in this PR make sure our user module works as expected on Sensu Go 5.21.0 that broke the user API.


tests/unit/modules/test_user.py
```diff
@@ -38,6 +41,28 @@ def test_password_is_invalid(self, mocker):
...
+        client.validate_auth_data.assert_called_once_with('user', 'pass')
+        client.put.assert_called_once()
+
+        path, payload = client.put.call_args[0]
+        assert path == '/path/reset_password'
+        assert payload['username'] == 'user'
+
+        # (tadeboro): We cannot validate the value without mocking the bcrypt.
+        # And I would rather see that our code gets tested by actually using
+        # the bcrypt rather than mocking it out. This way, the messa
```
mancabizjak, line 62:
```suggestion
        # the bcrypt rather than mocking it out. This way, the message
```

tests/unit/modules/test_user.py
```diff
@@ -38,6 +41,28 @@ def test_password_is_invalid(self, mocker):
             username='user', password='pass',
         ))
 
+    def test_password_is_invalid_newer_than_5_21_0(self, mocker):
```
mancabizjak, line 44: Nitpick (:
```suggestion
    def test_password_is_invalid_5_21_0_or_newer(self, mocker):
```

plugins/modules/user.py
```diff
@@ -189,6 +207,15 @@ def main():
...
     path = utils.build_core_v2_path(None, 'users', module.params['name'])
 
+    try:
+        if not HAS_BCRYPT and client.version >= "5.21.0":
+            module.fail_json(
+                msg=missing_required_lib('bcrypt'),
+                exception=BCRYPT_IMPORT_ERROR,
+            )
+    except errors.SensuError as e:
+        module.fail_json(msg=str(e))
```
mancabizjak, line 217: Just wondering if this part deserves a unit test. Maybe not so much for the `except` to cover errors stemming from `client.version`, but for the bcrypt dependency part - at least a test to ensure failure when we're missing the required library with 5.21.0 or newer? I would love to see something along the lines of `test_failure_on_missing_bcrypt_5_21_0_or_newer`.

tadeboro:
I added another test that simulates the scenario where we are trying to manage Sensu Go 5.22.something and the bcrypt library is missing. If the tests turn green, things should be ready to go.

PR #221, opened 2020-10-05
Add v1 API payload comparator and sync function

tadeboro:
Because Sensu Go's v1 API nests the real payload in the spec key, we need to perform comparison a bit differently.

To ease the process of syncing the resources, we also introduced a thin wrapper around the sync method that by default uses a custom v1 comparator and prepares the final result for us.


mancabizjak (approved):
Thank you @tadeboro, this will come in very handy. LGTM :+1:

Issue #219, opened 2020-09-29
Add modules for managing secrets providers


tadeboro:
We need to create three new modules for managing secrets providers:

 1. `secrets_provider_vault` for managing HashiCorp Vault providers
 2. `secrets_provider_env` for managing Env provider
 3. `secrets_provider_info` for retrieving information about available providers.

The general information about the secrets management is available [here](https://docs.sensu.io/sensu-go/latest/operations/manage-secrets/secrets-management/), provider references can be found [here](https://docs.sensu.io/sensu-go/latest/operations/manage-secrets/secrets-providers/), API documentation is [here](https://docs.sensu.io/sensu-go/latest/api/secrets/).

Why did we split the provider management into per-provider modules? Because different providers have completely different sets of arguments. And because adding a new provider to Sensu Go is not something that is done often, having provider-specific modules should not incur too much of a maintenance burden.

PR #223, opened 2020-10-07
Add secrets providers modules

mancabizjak:
Resolves https://github.com/sensu/sensu-go-ansible/issues/219.

This PR adds modules for managing/retrieving information about Sensu Go secrets providers:
* `secrets_provider_info`
* `secrets_provider_env`
* `secrets_provider_vault`

The vault provider module was a bit of a struggle due to Sensu Go API not behaving as documented; several of the supposedly required client spec fields did not seem to be required at all. In addition, if we omitted to specify optional fields when creating a Vault provider, the returned object did not contain those fields with their documented defaults. Instead, the default values for their respective types were returned.

We ended up making the following compromise: we require `address`, `token`, `version` and `tls.ca_cert` params for the `present` state, as these seem to be the absolutely neccessary ones for communicating with Vault. The module does not require `max_retries`, though it is supposedly required by the API. 

As for the default values for omitted parameters, we decided not to set any for the sake of simplicity, but also due to the aforementioned uncertainties.


tadeboro:
I did not read the code but requiring `tls.ca_cert` seems like an odd thing to do since passing http:// address is a valid choice (not smart or secure, but valid nontheless).

mancabizjak:
You're right, I got a bit carried away looking at the examples on the Sensu Go reference page :) `tls` is now optional - I updated tests accordingly and rebased.

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+        description:
+          - Path to the client certificate file.
+        type: str
+      client_key_file:
+        description:
+          - Path to the client key file.
+        type: str
+      common_name:
+        description:
+          - Common name of the Vault server.
```
tadeboro, line 87: I would assume this is common name of the client. And I say assume because I am not sure even after reading the Sensu and Vault docs (so much about documentation intuitiveness). But the server's common name has no use here, so I guess this is client-related value.

plugins/modules/secrets_provider_env.py
```diff
@@ -0,0 +1,94 @@
...
+  - For more information, refer to the Sensu Go documentation at
+    U(https://docs.sensu.io/sensu-go/latest/operations/manage-secrets/secrets-providers/).
+version_added: 1.6.0
+extends_documentation_fragment:
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.state
+seealso:
+  - module: secrets_provider_vault
+  - module: secrets_provider_info
+  - module: secret
```
tadeboro, line 36: Maybe we should also add the `secret_info` module here?

plugins/modules/secrets_provider_info.py
```diff
@@ -0,0 +1,95 @@
...
+  - Retrieve information about Sensu Go secrets providers.
+  - For more information, refer to the Sensu documentation at
+    U(https://docs.sensu.io/sensu-go/latest/operations/manage-secrets/secrets-providers/).
+version_added: 1.6.0
+extends_documentation_fragment:
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.info
+seealso:
+  - module: secrets_provider_env
+  - module: secrets_provider_vault
```
tadeboro, line 34: Maybe we can add `secret` and `secret_info` modules here?

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+        type: str
+
+seealso:
+  - module: secrets_provider_env
+  - module: secrets_provider_info
+  - module: secret
```
tadeboro, line 93: What about `secret_info`?

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+
+    current_client = current["spec"]["client"]
+    desired_client = desired["spec"]["client"]
+    # Sensu Go API returns 'agent_address' field in the client spec,
+    # but this field is not meant to be set via the providers API.
+    if utils.do_differ(current_client, desired_client, "agent_address", "tls"):
+        return True
+    # Sensu Go API returns 'insecure' and 'tls_server_name' fields in the
+    # tls spec. We ignore them, as they are not meant to be set via the
+    # providers API.
```
tadeboro, line 138: Can we make a note somewhere about this? Because this seems to me like a Sensu Go bug in either implementation or documentation. But in any case, being forced to find these things out via the trial-and-error method is not OK.

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+    return utils.do_differ(current_client["tls"], desired_client.get("tls") or {},
+                           "insecure", "tls_server_name")
+
+
+def _format_seconds(seconds):
+    # Sensu API returns the configured timeout as a string, for instance
+    # 30 -> '30s', 60-> '1m0s', 3600 -> '1h0m0s'.
```
tadeboro, line 145: Great work getting this mapping out of Sensu Go backend.

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+    if seconds < 60:
+        return "{0}s".format(seconds)
+    if seconds < 3600:
+        m, s = divmod(seconds, 60)
+        return "{0}m{1}s".format(m, s)
+    h, r = divmod(seconds, 3600)
+    m, s = divmod(r, 60)
+    return "{0}h{1}m{2}s".format(h, m, s)
```
tadeboro, line 153: Nitpick: maybe we can move the `h`, `m`, and `s` calculation to the top of the function (it is safe operation) and then conditionally return based on values of `h` and `m`? I had something like this in mind:

```suggestion
    h, r = divmod(seconds, 3600)
    m, s = divmod(r, 60)
    if h:
        return "{0}h{1}m{2}s".format(h, m, s)
    if m:
        return "{0}m{1}s".format(m, s)
   return "{0}s".format(seconds)
```

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def build_vault_provider_spec(params):
+    if params["state"] == "absent":
+        return {}
+
+    client = arguments.get_spec_payload(
+        params, "address", "token", "version", "max_retries")
```
tadeboro, line 161: Nitpick: can we move the `)` to the new line and terminate the call argument list with a `,`? My OCD gland hurts real bad right now ;)

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+    if params.get("tls"):
+        client["tls"] = arguments.get_renamed_spec_payload(params["tls"], dict(
+            ca_cert_file="ca_cert",
+            client_cert_file="client_cert",
+            client_key_file="client_key",
+            common_name="cname",
```
tadeboro, line 167: I am not sure about this renaming here. I wholeheartedly agree that upstream parameter names are less that ideal and do not convey all the information they should. But it still feels to me that keeping familiarity with the upstream API might be more important here. Using upstream docs for reference in vast majority of cases is something our users really like.

tests/integration/molecule/module_secrets_provider_vault/converge.yml
```diff
@@ -0,0 +1,277 @@
...
+        auth:
+          url: http://localhost:8080
+        name: my-vault
+      register: result
+      ignore_errors: true
+
+    - ansible.builtin.assert:
+        that:
+          - result is failed
+          - "'state is present but all of the following are missing' in result.msg"
```
tadeboro, line 47: **No action needed, just a remark.**

I tend to test this using unit tests lately because 1) they are much faster, and 2) I can pin down the behavior more. That second point is important for making sure we do not alter the API in a backward-incompatible way.

**Again, no action needed, just information for the future.**

tadeboro (changes requested):
Thank you, @mancabizjak, for this contribution. The only real issue I have with your implementation is the renaming of the TLS arguments. It feels to me that renaming them for the sake if readability is not worth breaking the "compatibility" with the upstream documentaion. But again, this is just my feeling and if you feel that the renaming is justified, just say so and we will leave it as is.

plugins/modules/secrets_provider_vault.py
```diff
@@ -0,0 +1,246 @@
...
+      - Timeout (in seconds) for connection to Vault server.
+    type: int
+  max_retries:
+    description:
+      - Maximum number of times to retry failed connections to Vault server.
+    type: int
+  rate_limit:
+    description:
+      - Maximum number of secrets requests for per second.
+    type: float
```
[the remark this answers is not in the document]
mancabizjak: I completely agree, which is the reason why I originally made it an integer, but then changed my mind at the last minute to align with upstream ;)

plugins/modules/secrets_provider_vault.py
mancabizjak: I was actually not sure whether this renaming would be a good thing or a bad one. Thanks for your input, we're now aligned with the upstream.

tests/integration/molecule/module_secrets_provider_vault/converge.yml
mancabizjak: Noted, thank you!

mancabizjak (commented):
Thank you for your comments @tadeboro. I think I applied all the required changes. Most importantly, the `tls` param names are now in sync with the upstream API. Along the way, I discovered one more TLS-related parameter (`ca_path`) that is returned by the Sensu providers API but should be ignored in the comparison of current vs. desired payloads, so I updated that as well.

PR #224, opened 2020-10-09
Add support for secrets to check, mutator, and pipe_handler modules

### LINKED ISSUE #166: Add support for secrets in check, handler and mutator definitions
Sensu Go 5.17.0 introduced a new *secret* resource that can be used when defining a check, handler, or mutator. Sensu Go Ansible modules currently do not support this attribute.

In order to support this, we need to add another parameter to the affected modules. So the plan of attack is:

 1. Add new documentation fragment for secrets parameter.
 2. Add argument specification for secrets parameter,
 3. Update modules to include the documentation fragment and add an additional parameter to arg spec construction.
 4. Add an additional parameter to the payload construction call.

tadeboro:
Fixes #166

### DIFF
--- a/plugins/doc_fragments/secrets.py
+++ b/plugins/doc_fragments/secrets.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2020, XLAB Steampunk <steampunk@xlab.si>
+#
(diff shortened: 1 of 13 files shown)


mancabizjak (approved):
I really like your work @tadeboro. I must say I didn't spot anything unusual, and have no further suggestions for improvements :) Thank you!

PR #227, opened 2020-10-29
Specify remote docker version in CI/CD configuration

tadeboro:
CircleCI started deprecating outdated stuff that is not supported anymore [1]. Fortunately, the only thing that affected us was the
docker version and that one was relatively easy to fix.

[1] https://discuss.circleci.com/t/old-linux-machine-image-remote-docker-deprecation/37572/1


tadeboro:
This approach increased our testing times too much, so we instead opted for the #229

PR #228, opened 2020-10-29
Switch to using Docker images from quay.io repositories

mancabizjak:
We recently migrated Docker images used in molecule integration tests from [DockerHub](https://hub.docker.com/u/xlabsi) to [Quay](https://quay.io/organization/xlab-steampunk) repositories. 

With this commit we ensure that Docker images are pulled from the new source. We also update the helper script for building Docker images so that images are pushed to the new location.


tadeboro (approved):
Thanks, @mancabizjak, for this PR. It is greatly appreciated.

PR #229, opened 2020-11-03
Transition to machine executor

tadeboro:
CircleCI started deprecating outdated stuff that is not supported anymore [1]. Fortunately, the only thing that affected us was the
docker version and that one was relatively easy to fix. But there is a nasty twist: when we specify docker version in our remote_docker step, our test times go up by a factor of 1.5 (20 minute runs take 30 minutes with the latest docker version).

We first tried to mitigate the increased test times by bumping the parallelism of our jobs, but unfortunately, this did not fix the issue
because the docker operations started to time out. And because a flaky test is worse than no test (it is waaaaay to easy to get accustomed to seeing errors in CI/CD console and stop reacting to them), we abandoned that approach.

The approach that did work for us was to get rid of the remote docker altogether and use machine executor to run our integration tests. They are a bit slower compared to the state before the change (compared to the old and deprecated docker version), but because we can safely bump the parallelism to a higher value now, the complete test suite actually executes faster.

Now all we need to do is get automatic timing of our tests and we should be golden. But that is something for another day.

[1] https://discuss.circleci.com/t/old-linux-machine-image-remote-docker-deprecation/37572/1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,39 +107,39 @@ jobs:
                 path: tests/output/junit
 
   integration_test_git:
+    parallelism: 6
     parameters:
       ansible:
         description: Ansible version to use
         type: string
-    docker:
-      - image: "circleci/python:3.8"
+    machine:
+      image: ubuntu-2004:202010-01
     working_directory: ~/ansible_collections/sensu/sensu_go
-    parallelism: 4
     steps:
       - wrapper:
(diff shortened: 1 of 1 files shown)


mancabizjak (approved):
Looks good :+1: Thank you @tadeboro for digging into this and not taking the easy way out.

PR #233, opened 2020-11-16
Enabled self-timing on integration tests

tadeboro:


mancabizjak (approved):
Thank you @tadeboro, I have nothing to add except that l really like the results.

PR #235, opened 2020-12-03
Remove CentOS 6 integration tests

tadeboro:
CentOS 6 reached its end of live on November 30. Because base repos were disabled at that time, it is now no longer feasible to test against the CentOS 6 distribution.


mancabizjak (changes requested):
Thanks @tadeboro. We should also update the "tested platforms" section in doc files for all the roles, since CentOS 6 is still mentioned there.

tadeboro:
Ugh, I totally forgot about the documentation. I knew I need a second set of exes on this PR. Thanks, @mancabizjak!

mancabizjak (approved):

PR #248, opened 2021-01-10
Coc updates

tadeboro:
Changes in this PR do two things:

 1. Add contact information to CoC.
 2. Generalize CoC a bit since the previous version was a bit too focused on conferences and other in-person events.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,6 +1,39 @@
-# Code of Conduct
+# Overview
 
 > Online version of this document is available at https://sensu.io/conduct
+
+As a member of the Sensu community, we ask that you keep the following in mind:
+
+  * Be kind and welcoming to others.
+  * Respect personal space and ask before you enter it. Remember not
+    everyonelikes to be touched or hugged or joked with.
+  * Be mindful of language. Do not disparage others. Listen more than you talk.
+  * Remember that harassment and sexist, racist, comments or behavior are not
+    appropriate in our Community.
+
+Sensu in-person and online events are intended for networking and collaboration
+in the monitoring community. Attendees violating these rules may be asked to
+leave any event we host without a refund at the sole discretion of the event
+organizers. Thank you for helping make this a welcoming place for all.
+
+
+## Important numbers
+
+Conference staff is happy to help participants contact venue security or local
+law enforcement, provide escorts, or otherwise assist those experiencing
+harassment to feel safe for the duration of the event. We value your
+attendance.
+
+If your physical safety or the physical safety of others is at risk, please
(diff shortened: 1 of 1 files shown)


tadeboro:
@jillr @cybette @mscherer I did my best to generalize the CoC a bit and align it with requirements from https://github.com/ansible-collections/ansible-inclusion/discussions/6#discussioncomment-270218.

Since I never authored a CoC before, I added examples based on what we already practice when working on Sensu Go Ansible Collection (being patient with newcomers, saying thank you when contributions are accepted). But feel free to suggest corrections and clarifications.

Thank you all in advance.

CODE_OF_CONDUCT.md
```diff
@@ -1,13 +1,51 @@
...
+    appropriate in our Community.
+
+Sensu in-person and online events are intended for networking and
+collaboration in the monitoring community. Attendees violating these rules may
+be asked to leave any event we host without a refund at the sole discretion of
+the event organizers.
+
+The same rules apply when participating in other Sensu-related forums and
+mailing lists, contributing to code and documentation, and in private
+correspondence.
```
mscherer, line 19: To be clear, that part is "private correspondence" is limited to Sensu related space ? (asking due to the Drupal 2017 controversy). 

Like, if I send you a email that would be a CoC violation, would it be covered by the CoC ?

CODE_OF_CONDUCT.md
```diff
@@ -1,13 +1,51 @@
...
+
+## Important numbers
+
+Conference staff is happy to help participants contact venue security or local
+law enforcement, provide escorts, or otherwise assist those experiencing
+harassment to feel safe for the duration of the event. We value your
+attendance.
+
+If your physical safety or the physical safety of others is at risk, please
+dial or text 911. For other needs, contact the relevant authorities below:
```
mscherer, line 32: While I do not have a good solution to that, 911 is unlikely to work outside of North America. (for example, in Europe, that's 112). Not sure how to handle the wonderful world of non standard numbers.

CODE_OF_CONDUCT.md
```diff
@@ -1,13 +1,51 @@
...
+
+**Code of Conduct team**  
+The Sensu Community team  
+community@sensu.io  
++1 855-997-3678 ext 510
```
mscherer, line 37: I think a tel: url could make giving a phone call easier, just click from you phone.

To be clear also, since that would also apply to meetups (once they will be back), is the line 24/7 ? 
And I also guess that this requires to speak english (which is obvious since the text is in english, but might be lost if the CoC is translated later, or if people copy the number somewhere else)

mscherer:
Seems globally good to me.

PR #256, opened 2021-02-22
Remove unused source repositories

amdprophet:
Signed-off-by: Justin Kolberg <amd.prophet@gmail.com>

Removes deb/rpm source repositories. We don't (and probably never will) make use of the source repositories.


tadeboro:
Hi @amdprophet. We kept those in because this is what packagecloud script installs when adding repositories. We can of course remove them if they are not needed, I just wanted to clarify why they are here today.

So, do you still want them gone, @amdprophet?

amdprophet:
They likely won't get used but let's keep them there for parity with the packagecloud script.

PR #257, opened 2021-02-22
Make allow_downgrade configurable for dnf/yum

amdprophet:
Signed-off-by: Justin Kolberg <amd.prophet@gmail.com>

We have a need to downgrade packages internally. This makes `allow_downgrade` a configurable option (defaults to `no`).

### DIFF
--- a/roles/install/defaults/main.yml
+++ b/roles/install/defaults/main.yml
@@ -6,7 +6,6 @@ packagecloud_auth: ""
 channel: stable
 version: latest
 build: latest
-allow_downgrade: no
 components:
   - sensu-go-backend
   - sensu-go-agent
--- a/roles/install/defaults/main.yml
+++ b/roles/install/defaults/main.yml
@@ -6,6 +6,7 @@ packagecloud_auth: ""
 channel: stable
 version: latest
 build: latest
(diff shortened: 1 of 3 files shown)


amdprophet:
I'm not sure if a similar change is needed for `apt` & `win_package` -- I didn't see any mention of allowing downgrades in the Ansible documentation.

tadeboro:
Hmm, IIRC, apt and windows already allow downgrading packages. So maybe we do not even need yet another variable and just "hardcode" the permission to downgrade directly into yum and dnf tasks?

tadeboro (approved):
Thank you, @amdprophet, for your contribution!

wat-usmdon:
FYI

It doesn't look like apt allows this at all. I got this message because I have an apt upgrade in my playbook and I didn't mark the sensu-go-backend package as held.

```
Reading package lists...
Building dependency tree...
Reading state information...
The following packages will be DOWNGRADED:
  sensu-go-backend
0 upgraded, 0 newly installed, 1 downgraded, 0 to remove and 0 not upgraded.
E: Packages were downgraded and -y was used without --allow-downgrades.
```

https://github.com/ansible/ansible/pull/33677

PR #276, opened 2021-03-30
Remove hard dependency on ansible.windows

### LINKED ISSUE #275: v1.9.1 onwards depends on ansible.windows which does not exist
**Bug description**
A clear and concise description of what the issue is.
There is a dependency on ansible.windows which does not exist on Automation Hub, hence its failing the building of execution environment
Also any customer who will try to install the collections from Automation Hub will also fail because of the unavailability of the collection, please fix this asap

cc: @tadeboro

tadeboro:
Because the `ansible.windows` Ansible Collection is not certified yet, hard dependency on that collection will prevent Automation Hub users from installing the Sensu Go Ansible Collection. We also added some additional instructions on how to manually install the Windows Ansible Collection for those users.

Do note that removing the `ansible.windows` dependency from the metadata has one major drawback: execution environments will not be able to manage Windows agent installations.

Fixes #275

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -78,3 +78,17 @@ Go collection by running::
 
 .. _releases page:
    https://github.com/sensu/sensu-go-ansible/releases
+
+
+Installing the Windows Ansible Collection
+-----------------------------------------
+
+If we are using Ansible Base or Ansible Core, we need to install the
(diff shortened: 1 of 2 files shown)


gberginc (approved):
Approving for the time being for this certified collection to be included in execution environment.

PR #284, opened 2021-04-09
Add auth provider info module

### LINKED ISSUE #278: Add auth_provider_info module
This module should be able to fetch exiting authentication providers. The API access point that we should use is documented at https://docs.sensu.io/sensu-go/latest/api/authproviders/#authproviders-get

We did something similar with the secrets providers already. The [secrets_provider_info](https://github.com/sensu/sensu-go-ansible/blob/master/plugins/modules/secrets_provider_info.py) module can serve as a template.

mdolin:
Resolves #278 
This module fetches existing authentication providers.

### DIFF
--- a/plugins/modules/auth_provider_info.py
+++ b/plugins/modules/auth_provider_info.py
@@ -1,0 +1,127 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2020, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["stableinterface"],
+    "supported_by": "certified",
+}
+
+DOCUMENTATION = """
+module: auth_provider_info
+
+author:
+  - Aljaz Kosir (@aljazkosir)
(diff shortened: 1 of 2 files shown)


plugins/modules/auth_provider_info.py
```diff
@@ -0,0 +1,127 @@
...
+      insecure: False
+      port: 636
+      security: 'tls'
+      trusted_ca_file: ''
+      user_search:
+        attribute: 'uid'
+        base_dn: 'dc=acme,dc=org'
+        name_attribute: 'cn'
+        object_class: 'person'
+    username_prefix: ''
```
tadeboro, line 90: This should be a list.

```suggestion
    - metadata:
        name: 'openldap'
      groups_prefix: ''
      servers:
        binding:
          password: 'YOUR_PASSWORD'
          user_dn: 'cn=binder,dc=acme,dc=org'
        client_cert_file: ''
        client_key_file: ''
        default_upn_domain: ''
        group_search:
          attribute: 'member'
          base_dn: 'dc=acme,dc=org'
          name_attribute: 'cn'
          object_class: 'groupOfNames'
        host: '127.0.0.1'
        insecure: false
        port: 636
        security: 'tls'
        trusted_ca_file: ''
        user_search:
          attribute: 'uid'
          base_dn: 'dc=acme,dc=org'
          name_attribute: 'cn'
          object_class: 'person'
      username_prefix: ''
```

tadeboro (changes requested):
Thank you, @mdolin for your contribution. I found one small error in the return value sample that we need to fix. Apart from this, the code looks great.

tadeboro (approved):

PR #285, opened 2021-04-14
Add LDAP authentication provider

### LINKED ISSUE #279: Add ldap_auth_provider module
This module should allow us to manage LDAP authentication providers. The Sensu Go documentation contains a [reference](https://docs.sensu.io/sensu-go/latest/operations/control-access/ldap-auth/) and [web API](https://docs.sensu.io/sensu-go/latest/api/authproviders/#authprovidersname-put).

The [secrets_provider_vault](https://github.com/sensu/sensu-go-ansible/blob/master/plugins/modules/secrets_provider_vault.py) can server as a template here.

mdolin:
Resolves #279
This module allows managing LDAP
authentication providers.

### DIFF
--- a/plugins/modules/ldap_auth_provider.py
+++ b/plugins/modules/ldap_auth_provider.py
@@ -1,0 +1,370 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+# Copyright: (c) 2020, XLAB Steampunk <steampunk@xlab.si>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+ANSIBLE_METADATA = {
+    "metadata_version": "1.1",
+    "status": ["stableinterface"],
+    "supported_by": "certified",
+}
+
+DOCUMENTATION = """
+module: ldap_auth_provider
+author:
+  - Aljaz Kosir (@aljazkosir)
+  - Manca Bizjak (@mancabizjak)
+  - Miha Dolinar (@mdolin)
+  - Tadej Borovsak (@tadeboro)
+
+short_description: Manage Sensu LDAP authentication provider
+
+description:
+  - Create, update or delete a Sensu Go LDAP authentication provider.
+  - For more information, refer to the Sensu Go documentation at
+    U(https://docs.sensu.io/sensu-go/latest/operations/control-access/ldap-auth/).
+
+version_added: 1.10.0
+
+extends_documentation_fragment:
+  - sensu.sensu_go.requirements
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.name
+  - sensu.sensu_go.state
+
+options:
+  servers:
+    description:
+      - An array of LDAP servers for your directory.
+      - Required if I(state) is C(present).
+    type: list
+    elements: dict
+    suboptions:
+      host:
+        description:
+          - LDAP server IP address.
+        required: true
+        type: str
+      port:
+        description:
+          - LDAP server port.
+        type: int
+      insecure:
+        description:
+          - Skips SSL certificate verification when set to true.
+        type: bool
+        default: false
+      security:
+        description:
+          - Encryption type to be used for the connection to the LDAP server.
+        type: str
+        choices: [ insecure, tls, starttls ]
+        default: tls
+      trusted_ca_file:
+        description:
+          - Path to an alternative CA bundle file.
+        type: str
+      client_cert_file:
+        description:
+          - Path to the certificate that should be sent to the server if requested.
+        type: str
+      client_key_file:
+        description:
+          - Path to the key file associated with the client_cert_file.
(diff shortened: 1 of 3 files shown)


plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,345 @@
...
+EXAMPLES = """
+- name: Create a LDAP auth provider
+  sensu.sensu_go.ldap_auth_provider:
+    name: openldap
+    servers:
+      host: 127.0.0.1
+      group_search:
+        base_dn: dc=acme,dc=org
+      user_search:
+        base_dn: dc=acme,dc=org
```
tadeboro, line 172:
```suggestion
    servers:
      - host: 127.0.0.1
        group_search:
          base_dn: dc=acme,dc=org
        user_search:
          base_dn: dc=acme,dc=org
```

plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,345 @@
...
+                        type="str",
+                    ),
+                    binding=dict(
+                        type="dict",
+                        options=dict(
+                            user_dn=dict(
+                                type="str",
+                            ),
+                            password=dict(
+                                type="str",
```
tadeboro, line 267: We are missing `no_log=True` here because passwords should not be displayed in the output. This is also why sanity tests failed on 2.11.

plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,345 @@
...
+    path = utils.build_url_path(
+        API_GROUP, API_VERSION, None, "authproviders", module.params["name"]
+    )
+
+    payload = dict(
+        type="ldap",
+        api_version=API_VERSION,
+        metadata=dict(name=module.params["name"]),
+        spec=arguments.get_spec_payload(
+            module.params, "servers" "groups_prefix", "username_prefix"
```
tadeboro, line 331:
```suggestion
            module.params, "servers", "groups_prefix", "username_prefix"
```

plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,345 @@
...
+        supports_check_mode=True,
+        argument_spec=dict(
+            arguments.get_spec(
+                "auth",
+                "name",
+                "state",
+            ),
+            servers=dict(
+                type="list",
+                required=True,
```
tadeboro, line 232: Servers parameter is only required if `state: present`, so we need to handle this with a conditional. See https://github.com/sensu/sensu-go-ansible/blob/master/plugins/modules/asset.py#L161-L165 for example of how we can conditionally check this.

plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,345 @@
...
+  - sensu.sensu_go.requirements
+  - sensu.sensu_go.auth
+  - sensu.sensu_go.name
+  - sensu.sensu_go.state
+
+options:
+  servers:
+    description:
+      - An array of LDAP servers for your directory.
+    required: true
```
tadeboro, line 44: This is only required if `state: present`. See https://github.com/sensu/sensu-go-ansible/blob/master/plugins/modules/asset.py#L43-L47 for example of how we should handle this.

tadeboro (changes requested):
I found a few issues that should help get the code working. I did not test it, but I would assume that the code should be able to create new things now. Idempotency might be a bit problematic because dicts inside the `servers` list have optional values, but I am sure we will come up with a proper comparator for this without much problem.

plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,376 @@
...
+API_GROUP = "enterprise"
+API_VERSION = "authentication/v2"
+
+
+def _filter(payload):
+    # Remove keys with None values from dict
+    result = dict()
+
+    for key, value in payload.items():
+        if value is None:
```
tadeboro, line 228: This should read `if value is not None`.

plugins/modules/ldap_auth_provider.py
```diff
@@ -0,0 +1,376 @@
...
+            result[key] = value
+
+    return result
```
tadeboro, line 231: Shorter version:

```suggestion
    return dict((k, v) for k, v in payload.items() if v is not None)
```

tests/unit/plugins/modules/test_ldap_auth_provider.py
```diff
@@ -0,0 +1,333 @@
...
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (2, 7), reason="requires python2.7 or higher"
+)
+
+
+class TestFilter:
+    def test_filter(self):
+        payload = dict(a=1, b=None)
+
+        assert ldap_auth_provider._filter(payload) == dict(a=1)
```
tadeboro, line 31: We usually do not test private methods.

tadeboro (changes requested):
Apart from the lapsus in the filter function, things look OK and are almost ready to go.

tadeboro (approved):
Thank you, @mdolin!

PR #288, opened 2021-04-28
Hide param

mdolin:
Removes password and client_secret from return objects in auth_provider modules.

### DIFF
--- a/plugins/modules/ad_auth_provider.py
+++ b/plugins/modules/ad_auth_provider.py
@@ -208,7 +208,6 @@ object:
       default_upn_domain: 'example.org'
       binding:
         user_dn: 'cn=binder,dc=acme,dc=org'
-        password: 'YOUR_PASSWORD'
       group_search:
         base_dn: 'dc=acme,dc=org'
         attribute: 'member'
@@ -230,6 +229,15 @@ from ..module_utils import arguments, errors, utils
 
 API_GROUP = "enterprise"
 API_VERSION = "authentication/v2"
+
+
+def remove_item(result):
+    if result:
+        for server in result["servers"]:
+            if server["binding"] and "password" in server["binding"]:
+                del server["binding"]["password"]
+
+    return result
 
 
 def _filter(payload):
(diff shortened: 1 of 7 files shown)


plugins/modules/ad_auth_provider.py
```diff
@@ -232,6 +231,18 @@
...
 
 
+def remove_item(result):
+    if result:
+        for i in range(len(result["servers"])):
+            if (
+                result["servers"][i]["binding"]
+                and "password" in result["servers"][i]["binding"]
+            ):
+                del result["servers"][i]["binding"]["password"]
```
tadeboro, line 241: Maybe we should simplify this a bit:

```suggestion
        for server in result["servers"]:
            if server["binding"] and "password" in server["binding"]:
                del server["binding"]["password"]
```

plugins/modules/ldap_auth_provider.py
```diff
@@ -220,6 +219,18 @@
...
+def remove_item(result):
+    if result:
+        for i in range(len(result["servers"])):
+            if (
+                result["servers"][i]["binding"]
+                and "password" in result["servers"][i]["binding"]
+            ):
+                del result["servers"][i]["binding"]["password"]
+
+    return result
```
tadeboro, line 231: Same simplification applies as in AD auth provider.

plugins/modules/auth_provider_info.py
```diff
@@ -98,6 +98,34 @@
...
+                and "password" in result["servers"][i]["binding"]
+            ):
+                del result["servers"][i]["binding"]["password"]
+
+    return result
+
+
+def remove_client_secret(result):
+    if result and "client_secret" in result:
+        del result["client_secret"]
```
tadeboro, line 124:
```suggestion
        for server in result.get("servers", []):
            if server["binding"] and "password" in server["binding"]:
                del server["binding"]["password"]
                
        if "client_secret" in result:
            del result["client_secret"]
```

tadeboro (changes requested):
There are a few issues that we need to solve before the merge. And we should improve the commit messages while rebasing the PR.

tadeboro (approved):

PR #291, opened 2021-05-18
Fix in seealso string in auth provider info module

mdolin:
Fix for auth provider info module in "seealso" string.
We list other auth provider modules related to this one
as they were missing.


tadeboro (approved):
Thank you for this, @mdolin!

PR #303, opened 2021-09-24
Add bcrypt dependency to unit tests

tadeboro:
The latest version of ansible-test dropped quite a few dependencies and once such dependency is bcrypt. We should always ensure that that dependency is installed, but because it was already there, we forgot to add that piece of code to the test suite.

For reference, https://app.circleci.com/pipelines/github/sensu/sensu-go-ansible/2121/workflows/f753041a-c1ec-4151-a210-cf0b39b62e0b/jobs/10385 is how our CI failed on the devel version of ansible.


anzoman (approved):
Looks okay and I believe that the tests will now work. So, let's merge this one!

PR #304, opened 2021-10-01
Update CI setup for Ansible 2.13

tadeboro:
The Ansible development branch was bumped to 2.13, which means we need another ignore file. And while we were at it, we also added a test configuration file that allows us to skip testing on Python 2.6 on ansible-core >= 2.12.


anzoman (approved):
Everything looks good to me, @tadeboro thanks for this CI setup update.

PR #309, opened 2021-12-16
Update cluster_info.py

juremedvesek:
There is one obvious typo inside example "http:s//" instead of "https://". This PR fixes it.


justinc1 (approved):
Is ok.

PR #313, opened 2022-01-11
Unpin ansible-lint

tadeboro:
We pinned Ansible Lint a whle ago because we needed some urgent fixes
from the alpha version. Now that those fixes are part of the stable
release, we can remove the pin and start using latest stable version.


juremedvesek (approved):
Thanks.

PR #318, opened 2022-03-21
Improve module_utils/http.py python 3.5 compat

mkobel:
Fixes TypeError: the JSON object must be str, not 'bytes'
when running with python 3.5 on Debian 9

### DIFF
--- a/plugins/module_utils/http.py
+++ b/plugins/module_utils/http.py
@@ -33,6 +33,11 @@ class Response:
         if self._json is None:
             try:
                 self._json = json.loads(self.data)
+            except TypeError:  # Handle python 3.5 returning bytes
+                try:
+                    self._json = json.loads(self.data.decode('utf-8'))
+                except ValueError:
+                    self._json = None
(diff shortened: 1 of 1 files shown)


plugins/module_utils/http.py
```diff
@@ -33,6 +33,8 @@ def json(self):
         if self._json is None:
             try:
                 self._json = json.loads(self.data)
+            except TypeError: # Handle python 3.5 returning bytes
+                self._json = json.loads(self.data.decode('utf-8'))
```
juremedvesek, line 37: Would change block with code below, since we can fail in second case to. And also plese rebase to latest master, which fixes sanity tests.

```
            except TypeError:  # Handle python 3.5 returning bytes
                try:
                    self._json = json.loads(self.data.decode('utf-8'))
                except ValueError:
                    self._json = None
```

mkobel: @juremedvesek Thank you for your feedback. The pull request is rebased and updated.

juremedvesek (approved):

PR #319, opened 2022-04-12
Support ansible-core 2.14

uscinski:
### Summary

This pull request adds support for `ansible-core 2.14`.

### Reason for this pull request

Currently, the CI/CD process is failing. The CI/CD set for the project runs tests against the main development branch of Ansible, which is at the time of writing at version 2.14 and prior to this pull request has not been taken into account.

### Changes

Failing sanity tests were handled in a similar way than for `ansible-core 2.13`, i.e., by adding `ignore-2.14.txt` file, describing the checks that should be skipped. In addition to that, import of `distutils.version` module was substituted with `ansible.module_utils.compat.version`, keeping the former as a fallback for older Ansible versions.

### DIFF
--- a/plugins/module_utils/client.py
+++ b/plugins/module_utils/client.py
@@ -6,7 +6,10 @@
 from __future__ import absolute_import, division, print_function
 __metaclass__ = type
 
-from distutils import version
+try:
+    from ansible.module_utils.compat import version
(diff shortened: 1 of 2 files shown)


juremedvesek (approved):
LGTM

PR #342, opened 2023-03-31
Add symlink for AlmaLinux.yml for alma linux 9 support

aduston-wo:
It appears that in Alma Linux 9 the `ansible_distribution` variable has changed to `AlmaLinux`:

```
$ ansible -m setup localhost | grep distribution
[WARNING]: No inventory was parsed, only implicit localhost is available
        "ansible_distribution": "Rocky",
        "ansible_distribution_file_parsed": true,
        "ansible_distribution_file_path": "/etc/redhat-release",
        "ansible_distribution_file_variety": "RedHat",
        "ansible_distribution_major_version": "9",
        "ansible_distribution_release": "Blue Onyx",
        "ansible_distribution_version": "9.1",
```

Which results in the following error from the install playbook:

```
TASK [sensu.sensu_go.install : Include distro-specific vars (AlmaLinux)] *******
task path: /usr/share/ansible/collections/ansible_collections/sensu/sensu_go/roles/install/tasks/dnf/prepare.yml:2
Friday 31 March 2023  01:10:31 +0000 (0:00:00.101)       0:05:20.955 ********** 
fatal: [localhost]: FAILED! => {
    "ansible_facts": {},
    "ansible_included_var_files": [],
    "changed": false,
[...]

PR #344, opened 2023-04-03
File renaming for support

dnikoloski:


aduston-wo:
You too, eh? https://github.com/sensu/sensu-go-ansible/pull/342

I did this with a symlink because I wasn't sure if it would break backwards compatibility with older releases. I've only tested on AlmaLinux 9.1. Have you had a chance to test AlmaLinux 8?

PR #345, opened 2023-04-03
Adding bcrypt check

domendobnikar:
Check for bcrypt library and raise exception only where needed.

**Expected result**
The user is created/updated with the defined (bcrypt-ed) password hash.

- [x] Bcrypt check
- [x] Unit tests


juremedvesek (approved):
LGTM

PR #348, opened 2023-05-03
Update dead links

asachs01:
Updates links to point to correct paths in the Sensu docs

### DIFF
--- a/plugins/modules/event.py
+++ b/plugins/modules/event.py
@@ -25,7 +25,7 @@ short_description: Manage Sensu events
 description:
   - Send a synthetic event to Sensu.
   - For more information, refer to the Sensu documentation at
-    U(https://docs.sensu.io/sensu-go/latest/reference/events/).
+    U(https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/).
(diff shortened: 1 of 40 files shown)


plugins/modules/event.py
```diff
@@ -97,23 +97,6 @@
         description:
           - The total state change percentage for the check's history.
         type: int
-  metric_attributes:
```
juremedvesek, line 100: Is this attribute deprecated inside SensuGo? If it is, would like to know from which version. For compatibility reasons I I would leave it in, since it is not mandatory at all. In case that It must be removed, I will remove it also from argument specs.

juremedvesek (commented):
All but removing metrics argument looks ok. I opened discussion there.

plugins/modules/event.py
asachs01: @juremedvesek I got a little overzealous in removing that section. It hasn't been deprecated, but the link has moved to https://docs.sensu.io/sensu-go/latest/observability-pipeline/observe-events/events/#metrics-attribute.

asachs01: Should be good to go now.

